PR #53, opened 2026-02-18
Add cli reporter with spinner and verbose streaming output

engali94:


engali94:
GitHub bug, it doesn't show all commits in the branch

PR #67, opened 2026-02-20
Scaffold java backend

engali94:
Setting up the java backend so we can start adding actual codegen. Just the file structure for now, nothing functional yet.


boltffi_bindgen/src/render/java/templates/render_java/native.txt
```diff
@@ -0,0 +1,10 @@
+final class Native {
+    static {
+        System.loadLibrary("{{ module.lib_name }}");
+    }
```
ahmedre: this will break on vanilla java because there it's `{{ lib_name }}_jni`

ahmedre: do something similar to what you're doing in the Kotlin one:
https://github.com/boltffi/boltffi/blob/2ecf74f064f136a8796eb536d49d6364ab8974ca/boltffi_bindgen/src/render/kotlin/templates/render_kotlin/native.txt#L4

PR #71, opened 2026-02-21
Add java CLI support and experimental features system

engali94:


boltffi_cli/src/commands/pack.rs
```diff
@@ -392,6 +417,145 @@ fn pack_android(config: &Config, options: PackAndroidOptions, reporter: &Reporte
...
+        build_jvm_native_library(config, options.release, &step)?;
+        step.finish_success();
+    }
+
+    if options.regenerate {
+        let step = reporter.step("Generating C header");
+        run_generate_with_output(
+            config,
+            GenerateOptions {
+                target: GenerateTarget::Header,
```
osa8bit: this will reject configs when both Apple and Android are disabled for Java onl projects this will fail even when Java doesn't need the generic header

boltffi_cli/src/commands/pack.rs
```diff
@@ -392,6 +417,145 @@ fn pack_android(config: &Config, options: PackAndroidOptions, reporter: &Reporte
...
+        return Err(CliError::FileNotFound(jni_glue));
+    }
+    if !header.exists() {
+        return Err(CliError::FileNotFound(header));
+    }
+
+    let profile_dir = if release { "release" } else { "debug" };
+    let rust_lib = PathBuf::from("target")
+        .join(profile_dir)
+        .join(format!("lib{}.dylib", config.library_name()));
```
osa8bit: cargo normalizes hyphens to underscores in artifact names. A crate named `my-lib` produces `libmy_lib.dylib` not `libmy-lib.dylib`. it must use `config.library_name().replace('-', "_")` or the actual crate_name when specified.

PR #103, opened 2026-03-21
Make `bench-kotlin` work on Linux

bendk:
I ran into a few issues when running `just bench-kotlin` on my machine, It seemed like the main issue were the scripts didn't have linux support.  I feel like there's a better way to do some of this, but this makes things work on my machine at least.

* Updated scripts to work with `.so` files
* Use `include/linux` on linux
* Set the LD_LIBRARY_PATH when running benchmarks
* Added `gradle` dir. I was getting "Unable to access jarfile" errors pointing to the gradle directory.  I don't really understand gradle, but I think that dir needs to be checked in. I did a fresh install using `gradle wrapper` with Gradle 9.4.0.
* Run gradle clean before collecting the benchmarks, this ensures they're recalculated.  Is there a better way to do this?


engali94 (approved):
Sounds Good!

PR #124, opened 2026-04-01
feat: add dart generate command

nxmxsxs:


boltffi_cli/src/config.rs
```diff
@@ -87,6 +89,18 @@ pub struct TargetsConfig {
     pub java: JavaConfig,
 }
 
+#[derive(Debug, Serialize, Deserialize, Clone, Default)]
+pub struct LanguagesConfig {
```
engali94: Dart config should go under the existing targets section in Config, not a new languages section. Every language or platform in BoltFFI is a "target" (apple(swift), android(kotlin), java(android, jvm), wasm, etc.), so Dart should follow the same pattern. That keeps boltffi.toml consistent for users.

PR #134, opened 2026-04-03
Close build output channel after cargo exits

ahmedre:
run_command_streaming() kept an extra sender alive while draining
stdout/stderr, so the receiver loop could wait forever even after the
child process had exited. This caused boltffi commands to appear stuck
with no further output.

Drop the original sender once the reader threads are spawned and add a
regression test that verifies the streaming runner returns after the
child process exits.


engali94 (approved):
Nice catch!

PR #139, opened 2026-04-07
Desktop jvm and static linking Support

ahmedre:
- **Support explicit Android target architectures**
- **Support configurable Apple slice selection**
- **Fix check target resolution for disabled platforms**
- **Implement JVM host staticlib-first packaging**
- **Implement cross-host JVM packaging**
- **Handle Kotlin async callback completion explicitly**
- **Auto-load desktop JVM libs from packaged artifacts**


engali94 (approved):
NICE

PR #140, opened 2026-04-07
Support overlays for configuration

ahmedre:
It's useful to have different configurations for different use cases,
such as CI, builds, development, and so on. This patch allows passing in
a file via --overlay which overrides some of the base options set within
the boltffi.toml file.

### DIFF
--- a/boltffi_cli/src/main.rs
+++ b/boltffi_cli/src/main.rs
@@ -10,11 +10,11 @@ mod reporter;
 mod target;
 
 use clap::{Parser, Subcommand};
-use std::path::PathBuf;
+use std::{fmt, path::PathBuf};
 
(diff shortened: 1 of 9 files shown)


boltffi_cli/src/main.rs
```diff
@@ -510,24 +523,56 @@ fn execute_command(
...
 }
 
-fn load_config() -> Result<Config> {
+#[derive(Debug, Clone)]
+struct ConfigPaths {
+    overlay: Option<PathBuf>,
+}
+
+impl ConfigPaths {
+    fn from_cli(cli: &Cli) -> Self {
```
engali94: nit: we can use impl From<Cli> for ConfigPaths

engali94 (approved):

PR #145, opened 2026-04-08
Handle support for exported classes and callback trait exchange

laptou:
## Problem (MVP)

BoltFFI’s lowering pipeline treated **exported classes** (`#[boltffi::export] impl Type`) and **callback-shaped values** (`Box<dyn SomeCallbackTrait>`) like anonymous wire payloads: it had **no registry** for exported classes, return classification never produced **object/callback handle** strategies, and callback-trait trampolines assumed **wire buffers** for anything non-scalar. That broke real APIs that need:

- **Exported class by value** (e.g. `fn on_started(transfer: Transfer)`) — should cross as a **raw handle**, not `FfiBuf` / `WirePassable`.
- **Callback trait returning another callback** (e.g. `fn on_starting() -> Box<dyn OtherCallbacks>`) — should use **`CallbackHandle`** / wasm handle ids, not wire encoding.

Downstream, this showed up as missing `Passable`, `FfiSpan`/`FfiBuf` mismatches, and vtable signatures that could not normalize.

## What we changed

1. **Index**: scan `#[export] impl` and register **exported class** paths for the crate.
2. **Lowering**: classify **ObjectHandle** / **CallbackHandle** in shared return/param logic; move sync callback-return resolution under `lowering::returns` with a stable `exports::callback_return` re-export.
3. **ABI rules**: `ffi_rules` — callback vtable returns for handles use **out-parameter** writes (aligned with generated trampolines).
4. **Exports**: `ffi_class` emits **`Passable`** for the type (`*mut T` pack/unpack); exports/functions/records thread the new lowering.
5. **Trait export**: correct **`uses_wire_payload`** (wire-buffer path vs handle/scalar); **`Passable` for `Box<dyn Trait>`** (native `CallbackHandle` vs wasm `u32`); local/native/wasm trampolines handle params and returns accordingly.
6. **Bindgen**: refresh TS preamble snapshot (triple-slash reference line).

## How to validate

- `cargo test -p boltffi_macros --lib`
- `cargo test -p boltffi_tests`
- Downstream compile check (example): `cargo check --manifest-path path/to/core/Cargo.toml --features native`

## Commits

Seven small commits on `feature/handle-support-callback-classes` for reviewability (index → lowering → params → ffi_rules → exports → trait export → bindgen snapshot).


engali94:
thi seems to be stale pr we can re-open if it is still needed

PR #149, opened 2026-04-10
feat(dart): initial build and packaging support

nxmxsxs:
# Configurable platform architectures
Users can now provide select targets for ios, android and linux:
```toml
...
[targets.dart]
native_architectures = ["android:arm64", "ios:arm64", "ios_sim:x86_64", "linux:x86_64", "macos:arm64"]
```
An empty list implies the platform will be excluded during `boltffi [build,pack]`

# Packaging
`boltffi pack` generates a self contained dart package which the user can add as a dependency.
```
{{ crate_name }}
    ├── hook
    │   └── build.dart
    ├── lib
    │   └── {{ crate_name }}.dart
    ├── native
    │   ├──  {{ target_triple }}
    │        └── lib{{ crate_name }}.so
    └── pubspec.yaml
```
- `{{ crate_name }}/lib/{{ crate_name }}.dart`: contains the rendered dart bindings
- `{{ crate_name }}/hook/build.dart`: Adapted from [native_toolchain_rust](https://github.com/GregoryConrad/native_toolchain_rust). It's a build hook which packages the built library with the dart/flutter application. At the moment, only release builds of the library are supported.
- `{{ crate_name }}/native/{{ target_triple }}/lib{{ crate_name }}.so`: built library


boltffi_cli/src/commands/pack.rs
```diff
@@ -2203,6 +2213,47 @@ fn strip_cargo_target_selectors(cargo_args: &[String]) -> Vec<String> {
     filtered
 }
 
+fn pack_dart(config: &Config, options: PackDartOptions, reporter: &Reporter) -> Result<()> {
+    if !config.is_java_jvm_enabled() {
```
engali94: here you meant `config.is_dart_enabled()`

boltffi_bindgen/src/render/dart/emit.rs
```diff
@@ -45,7 +54,24 @@ impl DartEmitter {
...
+                version: None,
+                repository: None,
+            }
+            .render()
+            .unwrap(),
+            lib: output,
+            build: BuildHookTemplate {
+                crate_name: package_name,
+                crate_path: crate_dir,
+                build_mode: "release",
```
engali94: this is hardcoded that means the default flow builds debug artifacts and then emits a package whose hook only looks in `target/<triple>/release/...`, so the generated package will be broken by default
Better option would be extending dart config to choose the release mode or the hook needs to use the actual selected Cargo profile

boltffi_cli/src/commands/build.rs
```diff
@@ -9,6 +9,7 @@ pub enum BuildPlatform {
     Apple,
     Android,
     Wasm,
+    Dart,
```
engali94, line 13: boltffi pack all --experimental still still only dispatches Apple/Android/WASM/Java so it will skip Dart entirely

boltffi_bindgen/src/render/dart/emit.rs
nxmxsxs: Right. Unfortunately, from what I've researched, there doesnt seem to be a builtin way to pass down build mode automatically. 
The user would have to manually buiild the library using `boltffi build dart [--release]` and also we would have to provide a build mode option in the build hook code which the user would also have to manually input.
Solutions:
  -  One option, which was suggested [here](https://github.com/dart-lang/native/issues/3237) would be use an external build tool which would generate some metadata file containing the build mode + call the appropriate dart/flutter cli command which in turn would read from the metadata file to get the build mode.
  - Another option might be to have the build hook control the build process.
  This would required extending boltffi cli to build for specific platforms+architectures (`boltffi build --crate-path=... dart --architectures=android:arm64,linux:x86_64,...`). This way the build hook can selectively choose what to build for.

engali94: Hey @nxmxsxs, thanks for all the work on this the architecture configuration, the Linux plumbing, and the build/pack dispatch for Dart are all solid. There is one issue with the the distribution model,

Today across every language we already support, `boltffi` acts as a cross-compile-and-bundle toolchain. At pack time we pick a profile, we build the Rust library, and then we physically bundle the resulting binary into a distribution artifact alongside the generated bindings. The end user never touches Cargo, or never knows a `target/` directory exists, and doesn't need a Rust toolchain on their machine. for example:

- Swift's `pack_apple` builds the Rust lib, then `XcframeworkBuilder` copies the binaries into an `.xcframework`. Xcode consumes that framework opaquely.
- Android's `pack_android` builds per-architecture `.so`s, then `AndroidPackager` copies them into `jniLibs/`. Gradle consumes an AAR with the binaries already inside.
- Wasm's `pack_wasm` emits the `.wasm` (plus any wasm-bindgen glue) directly into the pack output. npm/bundlers pick it up as a static asset.
- Java/JVM has the same idea, native libs get bundled into the jar's resources.

In every case above we hand the consumer a self-contained package (binary + bindings glue) and the two worlds meet inside that package. The consumer's build system does not run Cargo.

the current Dart PR does  is that the generated Dart package is a different shape. Instead of bundling the binary, its `hook/build.dart` reaches back into the original crate's `target/<triple>/<profile>/` directory at the consumer's build time. That inverts our usual model, the Dart package is effectively a pointer to build output that lives elsewhere on disk, rather than a container of that build output.

I think the `native_toolchain_rust` adaptation pulled us toward this shape because that package is designed to build Rust from source on the consumer's machine, that's its whole point. But that's the opposite of what we want boltffi to do. We don't want Dart/Flutter users installing Rust toolchains, Android NDKs, and cross-compile linkers to consume a boltffi-generated package. We want them to `dart pub add <pkg>` (or path-depend on a pre-built package) and have everything Just Work, exactly like a Swift dev adding an XCFramework or an Android dev pulling an AAR.

The direction I'd suggest is restructuring `boltffi pack dart` to emit something like:

```
{{ crate_name }}/
├── hook/build.dart           # sth that resolves to ../native/<triple>/lib{{crate}}…
├── native/
│   └── <triple>/
│       └── lib{{ crate }}.{so,dylib,a}
├── lib/{{ crate_name }}.dart
└── pubspec.yaml
```

At pack-time we build the chosen profile and copy the resulting libs into `native/<triple>/` inside the package  same pattern as Apple copying into an XCFramework or Android into jniLibs. The hook resolves the lib relative to its own package (via `Platform.script` or the package root URI from `BuildInput`), never touches `target/`, and `build_mode` disappears from the template entirely because it's no longer meaningful in the package's world.

One caveat worth verifying before committing to this is I haven't double-checked that Dart's native_assets / hooks API cleanly supports resolving files from inside the hook's own package directory at build time (vs. requiring them to be declared as Dart assets in `pubspec.yaml`). If there's a platform constraint there, that changes the shape of the fix but even then, the principle holds, we want a bundled artifact, not a hook that points at the user's Cargo target dir.

And again most of this PR is keepers, the concern is really just the shape of what `pack dart` emits.

nxmxsxs: Fair points.

> I think the native_toolchain_rust adaptation pulled us toward this shape because that package is designed to build Rust from source on the consumer's machine, that's its whole point. But that's the opposite of what we want boltffi to do. We don't want Dart/Flutter users installing Rust toolchains, Android NDKs, and cross-compile linkers to consume a boltffi-generated package. We want them to dart pub add <pkg> (or path-depend on a pre-built package) and have everything Just Work, exactly like a Swift dev adding an XCFramework or an Android dev pulling an AAR.

Cool. The user would be able to add the package using `[dart|flutter] pub add 'pkg_name:{"path":"/path/to/pkg"}`.

> One caveat worth verifying before committing to this is I haven't double-checked that Dart's native_assets / hooks API cleanly supports resolving files from inside the hook's own package directory at build time (vs. requiring them to be declared as Dart assets in pubspec.yaml). If there's a platform constraint there, that changes the shape of the fix but even then, the principle holds, we want a bundled artifact, not a hook that points at the user's Cargo target dir.

Yes. BuildInput provides access to the package root.

engali94 (approved):
Well done!

engali94 (approved):

examples/platforms/dart/pubspec.lock
```diff
@@ -0,0 +1,420 @@
+# Generated by pub
+# See https://dart.dev/tools/pub/glossary#lockfile
+packages:
```
engali94: let us gitignore this one I guess we don't need it to be tracked

nxmxsxs:
@engali94 The current dart configurations for architectures feel a bit clunky to me. I was thinking instead we could defining an array option called `platform_targets` with values in the format `<platform>:<cpu-arch>` or similar. for example:
```boltffi.toml
[targets.dart]
platform_targets = ["android:arm64", "ios:arm64"]
```

Also what do you thinking of flipping the default to an empty list?

Alternatively. We could keep the current architectures configurationn structure but switch the defaults to empty lists + notify users during build/pack time if they dont provide targets.

engali94:
> @engali94 The current dart configurations for architectures feel a bit clunky to me. I was thinking instead we could defining an array option called `platform_targets` with values in the format `<platform>:<cpu-arch>` or similar. for example:
> 
> ```toml
> [targets.dart]
> platform_targets = ["android:arm64", "ios:arm64"]
> ```
> 
> Also what do you thinking of flipping the default to an empty list?
> 
> Alternatively. We could keep the current architectures configurationn structure but switch the defaults to empty lists + notify users during build/pack time if they dont provide targets.

 I agree that this is worth rethinking. The current Dart target configuration does feel clunky, and a single target-list model is probably easier to understand than several separate per-platform architecture fields.
My only suggestion would be to keep the implementation typed internally even if the config surface becomes more compact. My only hesitation with `platform_targets = ["android:arm64", "ios:arm64"]` is that I would not want to carry that around as raw strings internally. If we go  in that direction, I’d prefer to parse it into a typed Dart target representation and keep the rest of the pipeline strongly typed.

On the default, I’d be cautious about switching to an empty list by default. That would make the target effectively enabled but produce nothing unless explicitly configured, which may be more surprising than helpful.

boltffi_bindgen/src/render/dart/templates/render_dart/pubspec.yaml.txt
```diff
@@ -0,0 +1,14 @@
+name: {{ package_name }}
+
+environment:
+  sdk: ^3.10.8
+
+resolution: workspace
```
engali94, line 6: I tried this locally and it failed with “found no workspace root including it”

boltffi_bindgen/src/render/dart/templates/render_dart/hook.build.dart.txt
```diff
@@ -0,0 +1,135 @@
...
+
+    if (!input.config.buildCodeAssets) {
+      logger.info(
+        'buildCodeAssets is false; '
+        'skipping build of Rust code assets',
+      );
+      return;
+    }
+
+    const crateName = "{{ crate_name }}";
```
engali94: I think we’re mixing up two different names here, the Dart package name and the Rust artifact name. `pack_dart` resolves and copies native libraries using `config.crate_artifact_name()` but the generated build hook reconstructs the filename from `package.name`. That only works when those happen to be the same string. As soon as someone sets a different `[package].crate` or has a library target name that differs from the package name, the hook will look for the wrong native file.

boltffi_cli/src/commands/build.rs
```diff
@@ -67,6 +68,13 @@ pub fn run_build(config: &Config, options: BuildCommandOptions) -> Result<Vec<Bu
             println!("Building for wasm ({})...", profile);
             builder.build_wasm_with_triple(config.wasm_triple())?
         }
+        BuildPlatform::Dart => {
+            if !config.is_dart_enabled() {
```
engali94, line 72: This seems to skip the experimental gate for Dart.

boltffi_cli/src/commands/build.rs
```diff
@@ -79,6 +87,9 @@ pub fn run_build(config: &Config, options: BuildCommandOptions) -> Result<Vec<Bu
             if config.is_wasm_enabled() {
                 all_results.extend(builder.build_wasm_with_triple(config.wasm_triple())?);
             }
+            if config.is_dart_enabled() {
```
engali94, line 90: This seems to skip the experimental gate for Dart.

boltffi_cli/src/commands/build.rs
nxmxsxs: the build subcommand doesnt use the experimental flag. I assumed this was intentional as the build subcommand simply compiles the library for the selected targets.

engali94 (approved):
LGTM 🎉

nxmxsxs:
@engali94 Apologies. My recent merge of this pr pushed broken/invalid code. I dont exhaustively match on the `Platform` enum in  
  - https://github.com/boltffi/boltffi/blob/7774cffa798b79ac91fc1e2e3eb65aa8f25cf71b/boltffi_cli/src/pack/symbols.rs#L1222
  - https://github.com/boltffi/boltffi/blob/7774cffa798b79ac91fc1e2e3eb65aa8f25cf71b/boltffi_cli/src/pack/apple/mod.rs#L323

I believe this happened because I forgot to rebase and new code had been added.
I'm going to revert the merge.

Fix opened at #214

Issue #146, opened 2026-04-09
Support C#


mhedgpeth:
I'm building a WinUI3 desktop app with [crux](https://github.com/redbadger/crux) and would like to explore replacing our uniffi-based bindings with boltffi. In order to do that, I'd need C# support. C# would use P/Invoke (`[DllImport]`) to call the existing C ABI — no additional native bridge layer like JNI is needed. The scope would mirror how Java support was added (#54): start with a scaffold and primitives, then incrementally add type support through a series of PRs.

We were using `uniffi-bindgen-cs` but [this PR](https://github.com/NordSecurity/uniffi-bindgen-cs/pull/163) that keeps it maintained is not receiving attention, and [this ticket](https://github.com/NordSecurity/uniffi-bindgen-cs/issues/172) didn't get any response.

I'd like to follow the same incremental approach used during the Java implementation (with Java PRs linked below):

### Phase 1: Foundation (Peaceful)

- [x] Scaffold C# backend — `render/csharp/` module structure, register in `render/mod.rs` (ref: #67, branch: `feature/csharp-scaffold`)
- [x] CLI support + experimental flag — `GenerateTarget`, config, mark experimental (ref: #71, branch: `feature/csharp-cli`)
- [x] Update README — move C# from "Planned" to "In progress" (branch: `feature/csharp-readme`)

### Phase 2:  Code Generation (Normal)

- [x] Primitive function support (ref: #70, branch: `feature/csharp-primitives`) - #174 
- [x] String support — marshaling via P/Invoke (ref: #74, branch: `feature/csharp-strings`) - #180 
- [x] Records support — C# `struct` / `record struct` (ref: #76, branch: `feature/csharp-records`) - #183 
- [x] Enums — C-style → `enum`, data enums → abstract class + subclasses (ref: #87, branch: `feature/csharp-enums`) - #190 
- [x] Vec/list bindings (ref: #88, branch: `feature/csharp-vec`) - #201 
- [x] Option support — nullable types (ref: #89, branch: `feature/csharp-options`) - #205 
- [x] Refactoring to AST #246  
- [x] Classes — wrapping `IntPtr` handle with `IDisposable` (ref: #92, branch: `feature/csharp-classes`) - #255 

### Phase 3: Acceptance Testing (& in CI)

[...]

PR #150, opened 2026-04-10
Scaffold C# backend

mhedgpeth:
This PR implements the first step in #146 by creating the scaffolding for csharp, using java as a guide.


boltffi_bindgen/src/render/csharp/lower.rs
```diff
@@ -0,0 +1,40 @@
...
+impl<'a> CSharpLowerer<'a> {
+    pub fn new(ffi: &'a FfiContract, abi: &'a AbiContract, options: &'a CSharpOptions) -> Self {
+        Self { ffi, abi, options }
+    }
+
+    /// Walk the contracts and produce a C# module plan.
+    pub fn lower(&self) -> CSharpModule {
+        let lib_name = self
+            .options
+            .library_name
```
engali94, line 25: this might break the common Cargo case where the package name contains `-` because native library artifacts normalize that to _ we already ran into similar issue in java and now it already handles[ this in ](https://github.com/boltffi/boltffi/blob/23dd5075ee05ad18f76b831473a8a5e8dd4a5202/boltffi_bindgen/src/render/java/lower.rs#L215) 

for example a crate named `my-crate` would generate` [DllImport("my-crate")]` which will not resolve the built library unless users manually override `library_name`

engali94 (approved):
Well done @mhedgpeth! Thank you

boltffi_bindgen/src/render/csharp/lower.rs
mhedgpeth: Nice catch! I saw that was implemnted in Java, but not in Kotlin, so I added a standard way of handling that, updating Java, fixing the omission in Kotlin, and including it in C#.

I also added brief crate-level documents so we can make sure we're aligned, because it felt a little weird to put this in a `rules` crate but when I investigated it, that seemed like the right approach.

I also added `rstest` as a dev-dependency to make the unit testing easier.

I'll keep this unresolved to make sure you agree with the change; I wanted to try to eliminate the footgun altogether.

mhedgpeth:
This should be ready now with the above comment fixed, please take a look there and let me know if there are any concerns. Thanks!

boltffi_bindgen/src/render/csharp/lower.rs
engali94: Agree on all three. `boltffi_ffi_rules` is exactly the right place if both the macro and bindgen sides need to agree on what the library filename looks like, it belongs in that shared crate
for `rstest` is totally fine for parametrized tests. The `Name<LibraryName>` wrapper is a better solution than the inline `.replace` because it makes misuse a compile error instead of a runtime bug.

engali94 (approved):

PR #152, opened 2026-04-10
Fix Windows test failures caused by backslashes in generated TOML path

### LINKED ISSUE #151: cargo tests fail on Windows machines due to backslashes in generated TOML paths
On a windows machine, when you run cargo test on the project, there are errors on:

- `scan::tests::scan_demo_crate_includes_datetime_custom_type_exports`
- `scan::tests::scan_demo_crate_preserves_callback_return_type`
- `render::typescript::lower::tests::demo_callback_return_lowers_to_wrapped_function`

The tests need to be updated to normalize the path.

mhedgpeth:
This PR addresses #151 - `cargo test` now works on a windows machine by not using the double quoted string `"` which would interpret `C:\Users\...` as escaped strings.

closes #151


engali94 (approved):
Nice catch! thank you!

PR #153, opened 2026-04-10
Add uniffi-bindgen-java bench

skeet70:
I think better bench would be to change `kotlin-jvm-bench -> jvm-bench` and have all three so they're on the same JVM and such, but I didn't want to step on toes. Copied the Kotlin structure and infra mostly.

Also updates to uniffi 0.31, and JDK to LTS 25. `uniffi-bindgen-java` requires 22+ for FFM, and I didn't want the benches to knock the JNA Kotlin by excluding the 17 -> 25 JVM performance/JIT improvements.


engali94 (approved):
Thanks for adding it, thats great addition.
yeah a single `jvm-bench` module for all three backends would be much cleaner, let’s merge this first and track that consolidation as a follow up

Issue #155, opened 2026-04-12
Exported method with wire-decoded param and Vec return fails to compile


pperanich:
When an `#[export]` impl method takes a `#[data]` struct parameter (wire-decoded) and returns `Vec<T>` (which uses the `DirectVec` return strategy), the generated FFI wrapper has bare `return;` statements in error paths. Since the wrapper function returns `FfiBuf`, this is a type mismatch.

This is the same bug class as the constructor fix in PR #121, but affecting regular method exports.

PR #156, opened 2026-04-12
fix: DirectVec method early return generates valid FfiBuf

pperanich:
When an exported method takes a wire-decoded #[data] parameter and returns Vec<T> (DirectVec strategy), the error path for null pointer and decode failure generated bare `return;` instead of returning a valid FfiBuf. This caused a compile error:

    error[E0069]: `return;` in a function whose return type is not `()`

Fix: return `FfiBuf::default()` on error, matching the pattern used by other return strategies in the same match arm.

Same bug class as the constructor fix in PR #122, but for regular method exports.

Addressses #155.

### DIFF
--- a/boltffi_macros/src/lowering/returns/lower.rs
+++ b/boltffi_macros/src/lowering/returns/lower.rs
@@ -29,7 +29,7 @@ impl ResolvedReturn {
             }
             ValueReturnStrategy::Buffer(EncodedReturnStrategy::DirectVec) => {
                 quote! {
-                    return;
+                    return ::boltffi::__private::FfiBuf::default();
                 }
             }
             ValueReturnStrategy::Buffer(_) => match (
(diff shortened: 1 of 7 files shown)


boltffi_macros/src/lowering/returns/lower.rs
```diff
@@ -29,7 +29,7 @@ impl ResolvedReturn {
             }
             ValueReturnStrategy::Buffer(EncodedReturnStrategy::DirectVec) => {
                 quote! {
-                    return;
+                    return ::boltffi::__private::FfiBuf::default();
```
engali94: this would be the wrong return type for the wasm side, it needs to be platform-aware ex: `wasm return;`,  `native return FfiBuf::default();`
or even better we can split this helper into explicit wasm/native variants the same way callback returns already are

engali94 (commented):
Could you add a regression integration test in `boltffi_tests` for ex a `ClassTestFixture` method taking `FixturePoint` and `returning Vec<i32>` and also update the macro unit test to cover the wasm/native split for `DirectVec` invalid arg returns

boltffi_macros/src/lowering/returns/lower.rs
pperanich: Thanks for the feedback! I've addressed this now.

pperanich:
done

engali94 (approved):

PR #157, opened 2026-04-12
Generate record readers for blittable stream items

ahmedre:
This fixes the Kotlin benchmarks.


engali94 (approved):
nice!
could you also extend the tests to cover this case
and also update the demo to make sure it works end to end

ahmedre:
done

engali94 (approved):

PR #159, opened 2026-04-13
fix: emit useWireBytes helper unconditionally in Kotlin codegen

pperanich:
useWireBytes is a general-purpose helper that wraps a ByteArray into a little-endian ByteBuffer. It was incorrectly placed inside the `{%- if has_async_runtime %}` conditional block in the Kotlin preamble template, so it was only emitted when the module contained async functions.

Methods on exported classes that return Vec<u8> call useWireBytes regardless of whether async is used, causing "Unresolved reference 'useWireBytes'" compilation errors when the module has no async functions.

Move useWireBytes above the async_runtime conditional so it is always emitted.


engali94 (approved):
Nice catch thank you!

PR #162, opened 2026-04-13
feat: add WireEncode/WireDecode impls for Box<T> and ()

pperanich:
Adds `WireEncode`/`WireDecode` implementations for two previously-unsupported types: `Box<T>` and `()`.

### `Box<T>`

`Box<T>` currently has no wire impls. Any type containing a `Box<SomeWireType>` field fails to satisfy the derive bounds. This PR adds a wire-transparent implementation that delegates all trait methods to the inner `T`:

- `encode_to` writes the same bytes as bare `T`
- `decode_from` decodes as `T`, then boxes
- `ENCODING_KIND`, `wire_size`, `is_fixed_size`, `fixed_size` all inherit from `T`

This unlocks common patterns like `Option<Box<T>>`, `Vec<Box<T>>`, and `Result<Box<T>, E>` where boxing is used for indirection, sizing, or ownership reasons unrelated to the FFI boundary itself.

### `()` decode

`WireEncode for ()` already existed (zero-size encoding), but `WireDecode` was missing. This prevents `Result<(), Error>` and `Option<()>` from crossing the FFI boundary. The new decode impl reads zero bytes and returns `((), 0)`.

### Scope

This PR intentionally covers only the core wire impls. It does **not** attempt to deliver full support for recursive enum types like:

```rust
enum Tree {
    Leaf(i32),
    Node(Box<Tree>, Box<Tree>),
}
```

Recursive types depend on the full scan -> IR -> render pipeline producing correct `indirect enum` (Swift), `sealed class` (Kotlin), and equivalent shapes across all targets. They also raise concerns around stack-depth safety in the recursive wire encode/decode path and lack cycle detection. 

With the changes in this PR, `Box<T>` is useful for non-recursive patterns (`Option<Box<Large>>`, `Box<String>` fields, etc.) where the box is an ownership/sizing tool rather than a recursion enabler.

### Codegen compatibility

No changes needed in downstream codegen. The type scanner in `boltffi_bindgen/src/scan/mod.rs` already unwraps `Box<T>` (and `Arc<T>`) to the inner type during the scan phase, before any language-specific codegen runs. All 7 targets (Swift, Kotlin, Dart, TypeScript, Java, JNI, C) receive the unwrapped type and emit their native equivalent.

### Tests

25 unit tests covering both types:

- `Box<T>` wire-transparency (byte-identical to bare `T` for `i32`, `String`, `Vec<i32>`)
- `Box<T>` trait delegation (`ENCODING_KIND`, `is_fixed_size`, `fixed_size`)
- `Box<T>` roundtrips including `Box<Box<i32>>` (nesting, not recursive type definition) and `Option<Box<i32>>`
- `()` decode from empty and non-empty buffers
- Composite patterns: `Result<(), E>` (Ok/Err), `Option<()>`

### DIFF
--- a/boltffi_core/src/wire/decode.rs
+++ b/boltffi_core/src/wire/decode.rs
@@ -474,6 +474,141 @@ mod tests {
         assert_eq!(size, written);
     }
 
+    mod unit_type {
+        use super::*;
+
+        #[test]
+        fn unit_decode_from_empty_buffer() {
+            let buf: [u8; 0] = [];
+            let (value, consumed) = <()>::decode_from(&buf).unwrap();
+            assert_eq!(value, ());
+            assert_eq!(consumed, 0);
+        }
+
+        #[test]
+        fn unit_decode_from_nonempty_buffer() {
+            let buf = [0xFF; 8];
+            let (value, consumed) = <()>::decode_from(&buf).unwrap();
+            assert_eq!(value, ());
+            assert_eq!(consumed, 0);
+        }
+
+        #[test]
+        fn unit_roundtrip() {
+            let original = ();
+            let mut buf = [0u8; 4];
+            let written = original.encode_to(&mut buf);
+            assert_eq!(written, 0);
+
+            let (_decoded, consumed) = <()>::decode_from(&buf).unwrap();
+            assert_eq!(consumed, 0);
+        }
+
+        #[test]
+        fn result_ok_unit_roundtrip() {
+            let original: Result<(), i32> = Ok(());
+            let mut buf = [0u8; 16];
+            let written = original.encode_to(&mut buf);
+
+            let (decoded, consumed) = Result::<(), i32>::decode_from(&buf).unwrap();
+            assert_eq!(decoded, Ok(()));
+            assert_eq!(consumed, written);
+        }
+
+        #[test]
+        fn result_err_with_unit_ok_roundtrip() {
+            let original: Result<(), i32> = Err(99);
+            let mut buf = [0u8; 16];
+            let written = original.encode_to(&mut buf);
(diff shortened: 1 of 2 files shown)


engali94:
Thanks @pperanich this is a useful direction, and I think it is addressing two real edge cases:

- `()` decode support, which is needed for shapes like `Result<(), E>`
- `Box<T>` as a wire transparent wrapper, which is a natural Rust pattern and especially relevant when users need heap indirection

The only thing Id be careful about is the recursive `Box<Tree>` example in the description. That is a good motivation for the feature, but it is a larger end-to-end claim than just adding the core wire impls, since recursive boxed types also depend on the scan/lowering/codegen path across targets.

For example this Rust shape:
```rust
#[boltffi::data]
enum Tree {
    Leaf(i32),
    Node(Box<Tree>, Box<Tree>),
}

```

would need to map cleanly in generated target languages too, and that is not trivial everywhere. For example in Swift you would typically need an indirect enum shape such as:

```swift
indirect enum Tree {
    case leaf(Int32)
    case node(Tree, Tree)
}

```
and in Kotlin you would expect something like:
```kotlin
sealed class Tree {
    data class Leaf(val value: Int) : Tree()
    data class Node(val left: Tree, val right: Tree) : Tree()
}

```

and similarly for the other languages.

So the fix makes total sense, but can we scope the description down? Unless we have integration tests for the recursive case, I'd rather that be a separate PR.

pperanich:
This makes complete sense! I updated the description above.

I'll investigate the recursion issue in a follow-on PR. I need to do some thinking on what will be required for all codegen targets.

engali94 (approved):

PR #163, opened 2026-04-13
split generate command into typed language backends

osa8bit:


boltffi_cli/src/commands/generate/backend.rs
```diff
@@ -0,0 +1,221 @@
...
+
+    pub fn write_output(&self, output_path: &Path, source_code: impl AsRef<[u8]>) -> Result<()> {
+        std::fs::write(output_path, source_code).map_err(|source| CliError::WriteFailed {
+            path: output_path.to_path_buf(),
+            source,
+        })
+    }
+}
+
+pub trait GenerateBackend {
```
osa8bit: LanguageGenerator would be better fit here
and the implementers would use SwiftGenerator instead of SwiftBackend etc...

engali94 (approved):

nxmxsxs (approved):

PR #167, opened 2026-04-14
Add support for python primitives and free functions

osa8bit:
You can review this as a generation PR, not a packaging PR.  you can run `boltffi generate python --experimental` on the demo crate to test it then inspect the generated `__init__.py`, `__init__.pyi`, and `_native.c` 
the packaging PR will be next.


engali94 (commented):
never mind, duplicate

engali94 (approved):
LGTM

boltffi_bindgen/src/render/python/templates/render_python/native_module.c.txt
```diff
@@ -0,0 +1,249 @@
...
+    if ({{ function.function_pointer_name() }} == NULL) {
+        boltffi_python_unload_library();
+        PyErr_SetString(PyExc_ImportError, "failed to resolve native symbol {{ function.ffi_symbol }}");
+        return 0;
+    }
+{% endfor %}
+    return 1;
+}
+
+static PyObject *boltffi_python_initialize_loader(PyObject *self, PyObject *library_path) {
```
engali94, line 92: according to the official docs http://docs.python.org/3/howto/free-threading-python.html  this static method might become become a data race under Python 3.13t + free-threaded mode.
the global function pointers and the `if (handle != NULL)` idempotency check become a data race. not blocking, just  keep it in mind we'd need atomics or a `pthread_once` pattern later

boltffi_bindgen/src/render/python/templates/render_python/native_module.c.txt
```diff
@@ -0,0 +1,249 @@
...
+{% for primitive in used_scalar_types %}
+static int {{ primitive.parser_name() }}(PyObject *value, {{ primitive.c_type_name() }} *out) {
+{% if primitive.uses_bool_parser() %}
+    if (!PyBool_Check(value)) {
+        PyErr_SetString(PyExc_TypeError, "expected bool");
+        return 0;
+    }
+    *out = value == Py_True;
+    return 1;
+{% elif primitive.uses_signed_long_long_parser() %}
```
engali94, line 117: AFAIK in python TRUE, FALSE are INTs in disguise so `echo_i32(True)` is the same as `echo_i32(1)`
correct me if I am wrong but seems this is not handled properly here 
we should keep native python behavior as much as possible

engali94 (approved):

boltffi_bindgen/src/render/python/emit.rs
```diff
@@ -1,41 +1,160 @@
...
 impl PythonEmitter {
-    pub fn emit(module: &PythonModule) -> String {
-        ModuleTemplate { module }.render().unwrap()
+    pub fn emit(module: &PythonModule) -> PythonPackageSources {
+        let package_directory = PathBuf::from(&module.module_name);
+        let package_version_literal =
+            format!("{:?}", module.package_version.as_deref().unwrap_or("0.0.0"));
+        let native_extension_name_literal =
+            format!("{:?}", format!("{}._native", module.module_name));
+        let native_source_path_literal =
```
engali94, line 30: This assumes both rust and python have same escape sytanx. Rust's Debug trait produces quoted strings like "0.0.0" or "demo_lib._native" which happen to also be valid Python string literals.
So if module_name ever contained something like a unicode character, Rust Debug would emit `\u{...}` which is invalid Python syntax it would choke on the curly braces

boltffi_bindgen/src/render/python/templates/render_python/native_module.c.txt
osa8bit: will handle that next pr

boltffi_bindgen/src/render/python/emit.rs
osa8bit: will handle that next pr

boltffi_bindgen/src/render/python/templates/render_python/native_module.c.txt
osa8bit: will handle that next pr

PR #174, opened 2026-04-16
Add primitive function support to C# backend

mhedgpeth:
This addresses the next task in #146 - Primitive function support

I was careful to follow the patterns for Java (#70)/Python where appropriate.

One difference here: C# marshals bools as a 4 byte structure, so I have to decorate the bool type and method params to force it to marshal as a single byte.

C# also has native unsigned types and platform-sized integers (i.e. `usize` equivalent) with `nint`/`nuint` so every Rust primitive maps 1:1 without wrapper classes or boxing.

I added unit tests for this but the acceptance tests are saved for a later phase, but as I run these I am running the cli against the `csharp` backend for `examples/demo` and looking over the output to make sure it looks right.

### DIFF
--- a/boltffi_bindgen/src/render/csharp/plan.rs
+++ b/boltffi_bindgen/src/render/csharp/plan.rs
@@ -12,4 +12,68 @@ pub struct CSharpModule {
     pub lib_name: Name<LibraryName>,
     /// FFI symbol prefix (e.g., `"boltffi"`).
     pub prefix: String,
+    /// Top-level primitive functions. Used by both the public wrapper class
+    /// and the `[DllImport]` native declarations — C# P/Invoke passes
+    /// primitives directly, so one struct serves both layers.
+    pub functions: Vec<CSharpFunction>,
+}
+
+impl CSharpModule {
+    pub fn has_functions(&self) -> bool {
+        !self.functions.is_empty()
+    }
+}
+
+/// A primitive function binding. Serves double duty: the template uses `name`
+/// and C# types for the public static method, and `ffi_name` for the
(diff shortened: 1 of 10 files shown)


mhedgpeth:
@engali94 it looks like rust 1.95.0 has a new clippy that is broken. I'll create a PR for just that which will fix this build

mhedgpeth:
Once #175 is merged, I'll rebase this, the build will work, and this will be ready.

boltffi_bindgen/src/render/csharp/plan.rs
```diff
@@ -12,4 +12,68 @@ pub struct CSharpModule {
...
+    }
+}
+
+/// A parameter in a C# function.
+#[derive(Debug, Clone)]
+pub struct CSharpParam {
+    /// camelCase parameter name, keyword-escaped with `@` if needed.
+    pub name: String,
+    /// C# type (e.g., `"int"`, `"double"`, `"bool"`).
+    pub csharp_type: String,
```
engali94: this might benefit from an enum and usage becomes strongly typed like `func.return_type == CSharpType::Bool`

engali94 (approved):

boltffi_bindgen/src/render/csharp/plan.rs
mhedgpeth: great idea, I moved to that model

mhedgpeth:
@engali94 do you want to re-evaluate these follow up commits or do you want me to merge once I do the follow ups?

engali94 (approved):
All good, thank you!

PR #175, opened 2026-04-16
fix: new clippy warning which wants match to include if

mhedgpeth:
This fixes a clippy warning introduced in the new stable version [today](https://blog.rust-lang.org/2026/04/16/Rust-1.95.0/), 1.95.0.

The clippy warning wants the code to include an if clause to a match if that's possible, and that was possible in the scan.rs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -119,8 +119,8 @@ jobs:
       - uses: dtolnay/rust-toolchain@nightly
         with:
           components: miri
-      - run: cargo miri setup
-      - run: cargo miri test -p boltffi -p boltffi_tests
+      - run: cargo +nightly miri setup
+      - run: cargo +nightly miri test -p boltffi -p boltffi_tests
 
   cross-build:
     name: Cross Build (${{ matrix.os }})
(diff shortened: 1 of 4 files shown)


engali94:
right now CI is tracking the latest stable Clippy while contributors may be on an older local toolchain that means this exact pattern can be reintroduced later by someone who formats and validates locally and only fails once CI runs on a newer Rust release. I think we should pin the Rust toolchain with `rust-toolchain.toml` and have CI use that same pinned version. we might use a grace window, like N-1 or N-2 for the version, let us pint it to `1.94.1`

mhedgpeth:
> right now CI is tracking the latest stable Clippy while contributors may be on an older local toolchain that means this exact pattern can be reintroduced later by someone who formats and validates locally and only fails once CI runs on a newer Rust release. I think we should pin the Rust toolchain with `rust-toolchain.toml` and have CI use that same pinned version. we might use a grace window, like N-1 or N-2 for the version, let us pint it to `1.94.1`

Great I'll do that here

mhedgpeth:
OK @engali94 this now introduces the `rust-toolchain.toml`, the GHA are updated, and I created a related issue to take care of the policy you propose. I left the clippy change since this was a good change to make, to avoid future work. When you approve, I'll merge and then will rebase my C# change.

engali94 (approved):
Nice!

PR #179, opened 2026-04-16
Unify demo driven benchmark harnesses and add release benchmark publishing

engali94:
This PR cleans up the benchmark setup so it is not a pile of separate one off benchmark code anymore. The main change is that `examples/demo` is now the source of truth for the benchmarked Rust surface. I pulled the missing benchmark APIs into the right demo modules, added the benchmark overlay config, and then rewired the Swift, Kotlin/JMH, Java/JMH, and WASM benchmark harnesses to use that shared demo surface instead of drifting copies. 

The other half of the PR is the publishing path. Instead of storing benchmark JSON in the main repo, the release benchmark workflow now runs on tag push, executes the non-mobile benchmark suites on `macos-latest`, collects the normalized run documents, and publishes them into the benchmark archive repo. I added a real archive publisher script for that, and it treats archived raw runs as the source of truth and regenerates the derived views from there

The benchs dashboard: https://github.com/boltffi/benchmarks-dashboard

---

<table>
  <tr>
    <td></td>
    <td></td>
  </tr>
</table>


osa8bit (approved):

mhedgpeth:
@engali94 does my addition of the C# benchmark test affect this? If so I'm happy to include that here once I merge #183 - which should happen once the checks are finished, after I incorporated your feedback.

engali94:
> @engali94 does my addition of the C# benchmark test affect this? If so I'm happy to include that here once I merge #183 - which should happen once the checks are finished, after I incorporated your feedback.



> @engali94 does my addition of the C# benchmark test affect this? If so I'm happy to include that here once I merge #183 - which should happen once the checks are finished, after I incorporated your feedback.

Yes it needs to be adapted and included in the dashboard, I am still working on it so no problem once you merge I will pull it and adapt to my changes

mhedgpeth:
> Yes it needs to be adapted and included in the dashboard, I am still working on it so no problem once you merge I will pull it and adapt to my changes

Sounds good, I merged it just now, thanks!

PR #180, opened 2026-04-16
introduce C# strings and tests

mhedgpeth:
This is the latest in #146 - introduces strings.

This is also the point that things got quite abstract to look at so to maintain my sanity I added the acceptance tests for strings, so I can see that it works. I'll add these as I go.

### DIFF
--- a/boltffi_bindgen/src/render/csharp/templates/render_csharp/native.txt
+++ b/boltffi_bindgen/src/render/csharp/templates/render_csharp/native.txt
@@ -13,12 +13,22 @@
         internal static string ReadString(FfiBuf buf)
         {
             if (buf.ptr == IntPtr.Zero) return "";
+            nuint bufLen = (nuint)buf.len;
+            if (bufLen < 4) throw new InvalidOperationException("corrupt wire: buffer too small for length prefix");
             int len = Marshal.ReadInt32(buf.ptr);
             if (len < 0) throw new InvalidOperationException("corrupt wire: negative string length");
             if (len == 0) return "";
-            byte[] bytes = new byte[len];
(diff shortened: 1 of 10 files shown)


boltffi_bindgen/src/render/csharp/templates/render_csharp/native.txt
```diff
@@ -1,19 +1,43 @@
...
+
+    internal static class WireReader
+    {
+        internal static string ReadString(FfiBuf buf)
+        {
+            if (buf.ptr == IntPtr.Zero) return "";
+            int len = Marshal.ReadInt32(buf.ptr);
+            if (len < 0) throw new InvalidOperationException("corrupt wire: negative string length");
+            if (len == 0) return "";
+            byte[] bytes = new byte[len];
```
engali94: I think this might need bounds validation before reading from the memory in a way that doesn't hurt the performance if possibe
it reads the first 4 bytes as the string length and then copies `len` bytes from `buf.ptr + 4`, but it does not check that `buf.len >= 4` or that the declared payload fits inside `buf.len`

engali94 (approved):
Overall this looks nice, thank you for pushing it through.
I just left one comment on the string decode path, other than that, everything else looks fine to me.

boltffi_bindgen/src/render/csharp/templates/render_csharp/native.txt
mhedgpeth: Good catch, I spent some more time and added bounds checking, and also a better way to get the byte array copied.

engali94 (approved):

PR #183, opened 2026-04-17
Support C# Records

mhedgpeth:
This PR introduces records to the C# backend, as a part of #146 

Also as a part of this I evaluated whether an intra-`byte[]` reuse of `BinaryPrimitives` while taking on the `byte[]` copy performance hit would be better than a `Marshal`-based approach, and the latter ended up being ~25-30% faster, so that's what this goes with.

To test this, the PR also includes a C# benchmark harness, which I ran on my Mac M2 to generate the benchmark results, it is also nice to know that this all works on a Mac as well as a Windows machine, with `dotnet` installed.

This also avoids unsafe code which is an improvement over the `uniffi-bindgen-cs` crate.


boltffi_bindgen/src/render/csharp/lower.rs
```diff
@@ -54,57 +112,243 @@ impl<'a> CSharpLowerer<'a> {
...
+            binding_name,
+            bytes_binding_name,
+            param_name,
+            size_expr: emit::emit_size_expr(&encode_ops.size),
+            encode_expr,
+        })
+    }
+
+    fn abi_call_for_function(&self, function: &FunctionDef) -> Option<&AbiCall> {
+        use crate::ir::abi::CallId;
```
engali94: nit: let us move it tot he top of the module

engali94 (approved):
Looks good to me amazing work, a small note I think we should ignore the auto generated demo files

PR #186, opened 2026-04-17
Support recursive data enums

pperanich:
Builds on #162 to unlock recursive `#[boltffi::data]` enums across Swift, Kotlin, Java, and TypeScript, such as:

```rust
#[boltffi::data]
enum Tree {
    Leaf(i32),
    Node(Box<Tree>, Box<Tree>),
}
```

Swift was the blocker: its enums are value types, so a variant whose payload reaches back to the enclosing enum needs the `indirect` keyword or the generated code won't compile. Kotlin, Java, and TypeScript already handle recursion through reference-type hierarchies (sealed class, sealed interface, tagged union), but nothing was locking that in.

For Swift I went with per-variant `indirect case Node(...)` rather than marking the whole enum `indirect`, so non-recursive variants stay on the stack. Detection walks the `TypeExpr` graph with a visited set and follows `Option`, `Vec`, `Result`, transitive enum refs, and record fields, so cycles through other types are caught too.

Kotlin, Java, and TypeScript got no template changes, just snapshot and assertion tests to lock the existing shapes in. Added a `FixtureTree` to `boltffi_tests` plus a wire round-trip test in `boltffi_tests/tests/recursive_enum.rs`.

The Dart, C#, and Python backends don't render data-enum type declarations yet. C needs tagged-union support first. Decode depth limiting is worth addressing in core as a follow-up to prevent stack overflows.


engali94:
thanks @pperanich for the work I skimmed it quickly and I will do proper deep dive tomorrow
there is one thing I noticed `Box<T>::decode_from` is a direct recursive call which means a deep or malicious tree can now stackoverflow Rust or generated lang instead of returning a decode error, probably we need to set a limit on the nesting level depending on the size.

pperanich:
Yep! That is definitely a concern. I tried to indicate in the last line of the description above that we could address this in a future PR. Happy to work on this in this PR if desired though.

boltffi_bindgen/src/render/swift/lower.rs
```diff
@@ -614,6 +615,96 @@ impl<'a> SwiftLowerer<'a> {
...
+    }
+
+    fn type_contains_enum(
+        &self,
+        type_expr: &TypeExpr,
+        target: &EnumId,
+        visited_enums: &mut HashSet<EnumId>,
+        visited_records: &mut HashSet<RecordId>,
+    ) -> bool {
+        match type_expr {
```
engali94, line 660: I think this recursion check still misses custom type wrappers
our Swift backend emits unmapped custom types as plain typealiases so if a recursive edge is hiding behind something like `TreeAlias = Tree` this walk will miss it and we still end up generating a non-indirect recursive case

boltffi_bindgen/src/render/swift/lower.rs
```diff
@@ -614,6 +615,96 @@ impl<'a> SwiftLowerer<'a> {
...
     }
 
+    fn variant_is_recursive(&self, def: &EnumDef, index: usize) -> bool {
+        let EnumRepr::Data { variants, .. } = &def.repr else {
+            return false;
+        };
+        let Some(data_variant) = variants.get(index) else {
+            return false;
+        };
+        let mut visited_enums = HashSet::from([def.id.clone()]);
```
engali94, line 625: I’m a worried we’re rederiving recursion reachability in the Swift backend instead of consuming a shared answer from the IR and codec layer. It would be better to have this in shared place and consumed by backends

boltffi_bindgen/src/render/swift/templates.rs
```diff
@@ -902,6 +916,61 @@ mod tests {
...
+                            "right",
+                            "Tree",
+                            read_primitive(PrimitiveType::I32, offset("pos")),
+                            write_primitive(PrimitiveType::I32, "right"),
+                            None,
+                            None,
+                        ),
+                    ]),
+                    doc: None,
+                    is_recursive: true,
```
engali94, line 973: I dont think this snapshot is actually exercising the recursive field codec path. It locks the indirect case surface, but the left and right fields here are still wired with primitive read/write helpers, so this would still pass even if the real recursive lowering regressed

boltffi_bindgen/src/render/swift/lower.rs
```diff
@@ -614,6 +615,96 @@ impl<'a> SwiftLowerer<'a> {
...
+        let mut visited_records = HashSet::new();
+        self.payload_contains_enum(
+            &data_variant.payload,
+            &def.id,
+            &mut visited_enums,
+            &mut visited_records,
+        )
+    }
+
+    fn payload_contains_enum(
```
engali94, line 635: I think this should be a real traversal object rather than three positional mutable parameters passed through the whole walk. A `RecursionTraversal` type would make the ownership of target, visited_enums, and visited_records explicit

mhedgpeth:
Assigning to @pperanich to process feedback

boltffi_bindgen/src/render/swift/lower.rs
```diff
@@ -614,6 +615,96 @@ impl<'a> SwiftLowerer<'a> {
...
+                })
+            }
+            TypeExpr::Option(inner) | TypeExpr::Vec(inner) => {
+                self.type_contains_enum(inner, target, visited_enums, visited_records)
+            }
+            TypeExpr::Result { ok, err } => {
+                self.type_contains_enum(ok, target, visited_enums, visited_records)
+                    || self.type_contains_enum(err, target, visited_enums, visited_records)
+            }
+            _ => false,
```
ns-jjwathen: **Blocking:** this wildcard swallows `TypeExpr::Custom`. `CustomTypeDef.repr` is a full `TypeExpr` (a `custom_ffi` type's `FfiRepr` can be or contain this enum), and unmapped custom types render as plain typealiases in Swift — so recursion routed through one is still a value-type cycle, but this walk classifies it non-recursive and the generated enum omits `indirect` and fails to compile. Something like:

```rust
TypeExpr::Custom(id) => {
    // A native type mapping replaces the repr with an independent Swift type.
    if self.type_mappings.contains_key(id.as_str()) {
        false
    } else if let Some(def) = self.contract.catalog.resolve_custom(id) {
        self.type_contains_enum(&def.repr, target, visited_enums, visited_records)
    } else {
        false
    }
}
```

(add a visited set for custom ids if reprs can nest customs). While here, consider making the match exhaustive — listing `Void`/`Primitive`/`String`/`Bytes`/`Callback`/`Builtin`/`Handle` explicitly as `false` — so any future `TypeExpr` variant forces a decision here instead of silently bypassing detection.

boltffi_bindgen/src/render/swift/lower.rs
```diff
@@ -614,6 +615,96 @@ impl<'a> SwiftLowerer<'a> {
             .collect()
     }
 
+    fn variant_is_recursive(&self, def: &EnumDef, index: usize) -> bool {
```
ns-jjwathen: **Blocking (coverage):** nothing exercises this function — the only `is_recursive: true` in the PR is hand-set in the `templates.rs` snapshot test, so a regression here (even `return false`) passes the entire suite. This file already has a `mod tests` (line ~2446) that builds contracts and runs the real lowerer; please add cases asserting `SwiftVariant.is_recursive` for: direct self-recursion, recursion via `Option`/`Vec`/`Result`, mutual recursion A→B→A (both marked), recursion through a record field, recursion through a custom type, and a non-recursive variant that references a *different* self-recursive enum (must be `false`).

boltffi_bindgen/src/render/swift/lower.rs
```diff
@@ -614,6 +615,96 @@ impl<'a> SwiftLowerer<'a> {
...
+                def.fields.iter().any(|field| {
+                    self.type_contains_enum(
+                        &field.type_expr,
+                        target,
+                        visited_enums,
+                        visited_records,
+                    )
+                })
+            }
+            TypeExpr::Option(inner) | TypeExpr::Vec(inner) => {
```
ns-jjwathen: Nit: `Vec` doesn't actually need to be treated as a recursive edge for Swift — `Array` storage is heap-allocated, so `case children([Tree])` has finite inline size and compiles without `indirect`. Following it just adds a redundant box. `Option`/`Result` genuinely need it. Fine to keep as deliberate conservatism, but worth a comment either way.

boltffi_bindgen/src/render/kotlin/templates.rs
```diff
@@ -739,6 +739,53 @@ mod tests {
...
+                KotlinEnumVariant {
+                    name: "Node".to_string(),
+                    tag: 1,
+                    fields: vec![
+                        KotlinEnumField {
+                            name: "left".to_string(),
+                            kotlin_type: "Tree".to_string(),
+                            wire_decode_expr: "Tree.decode(reader)".to_string(),
+                            wire_size_expr: "left.wireEncodedSize()".to_string(),
+                            wire_encode: "left.encode(wire)".to_string(),
```
ns-jjwathen: This expr (and `right.encode(wire)` below) is fabricated: the rendered sealed class only defines `wireEncodedSize()`/`wireEncodeTo(wire)` and doesn't implement `WireCodable`, and the real lowerer emits `left.wireEncodeTo(wire)` (`kotlin/emit.rs:327`) — so the committed snapshot is Kotlin that would not compile, and a regression in real recursive-field lowering wouldn't fail this test. Please use the exprs the lowerer actually produces (or drive the fixture through the lowerer) and regenerate the snapshot with `cargo insta`. Same fabricated-expr issue in the Swift snapshot test (`reader.readI32()` for `Tree` fields vs. real `Tree.decode(from: &reader)`), the TS one (`writer.writeI32(v.left)`), and the Java test's `_v.left().encode(wire)`.

boltffi_tests/tests/recursive_enum.rs
```diff
@@ -0,0 +1,70 @@
+use boltffi::__private::wire::{WireDecode, WireEncode};
```
ns-jjwathen: Minor: all four tests are happy-path `.unwrap()` round-trips. Since nested decode is the new behavior, it'd be worth pinning that inner failures surface as `Err` rather than panic: (a) a `Node` whose right-subtree bytes are truncated, and (b) a nested payload carrying an out-of-range tag. That also documents intended untrusted-input behavior ahead of the deferred depth-limit work.

ns-jjwathen (commented):
_Review withdrawn — posted here in error (automation pointed at the wrong PR list). Apologies for the noise._

PR #188, opened 2026-04-18
Add support for python c style enums

osa8bit:
Also re organized the python modules


boltffi_bindgen/src/render/python/templates/render_python/native_module.c.txt
```diff
@@ -515,6 +520,188 @@ static int {{ primitive.vector_parser_name() }}(PyObject *value, boltffi_python_
...
+}
+
+static PyObject *boltffi_python_box_registered_enum(
+    PyObject *type_object,
+    PyObject *boxed_value,
+    const char *enum_name
+) {
+    PyObject *enum_value = NULL;
+    if (!boltffi_python_expect_registered_enum(type_object, enum_name)) {
+        Py_DECREF(boxed_value);
```
engali94, line 580: boxed_value can be NULL here and thus UB, probably want `Py_XDECREF` or just bail early if the arg is NULL before doing anything else.

boltffi_bindgen/src/render/python/primitives/cpython.rs
```diff
@@ -324,35 +339,51 @@ impl CPythonParameterExt for PythonParameter {
...
             }],
-            PythonType::Sequence(_) => vec![CPythonCBinding {
-                c_type_name: "boltffi_python_buffer_input".to_string(),
-                name: self.parser_state_name(),
-            }],
+            PythonType::Sequence(PythonSequenceType::Bytes)
+            | PythonType::Sequence(PythonSequenceType::PrimitiveVec(_))
+            | PythonType::Sequence(PythonSequenceType::CStyleEnumVec(_)) => {
+                vec![CPythonCBinding {
+                    c_type_name: "boltffi_python_buffer_input".to_string(),
```
engali94, line 356: these `boltffi_python_buffer_input` locals get declared at the top uninitialized, and `cleanup:` releases all of them. if arg parsing bails on p2 before p1's buffer gets populated we're handing garbage pointers to the release fn.
this can be fixed by just slapping = {0} on the decl

boltffi_bindgen/src/render/python/templates/render_python/native_module.c.txt
```diff
@@ -515,6 +520,188 @@ static int {{ primitive.vector_parser_name() }}(PyObject *value, boltffi_python_
...
+        Py_DECREF(sequence);
+        PyErr_SetString(PyExc_OverflowError, "sequence is too large for Rust");
+        return 0;
+    }
+    if ((size_t)sequence_length > ((size_t)UINTPTR_MAX - sizeof(int32_t)) / sizeof(int32_t)) {
+        Py_DECREF(sequence);
+        PyErr_SetString(PyExc_OverflowError, "sequence is too large for Rust");
+        return 0;
+    }
+    encoded_size = sizeof(int32_t) + ((size_t)sequence_length * sizeof(int32_t));
```
engali94, line 675: primitive vectors set `out->len` to the element count, but here we're setting it to encoded_size (byte count). not a bug if the consumer side agrees, but it's gonna bite someone later. worth a comment at minimum, or normalize to element count like the primitives do.

engali94 (approved):
nice reog btw 
I left few comments other than that all good

PR #189, opened 2026-04-18
Make the csharp demo and dotnet benchmarks runnable again

engali94:
the benchmarks now available at https://boltffi.github.io/benchmarks-dashboard/?machine=local-darwin-arm64-apple-m3-max&language=csharp

PR #190, opened 2026-04-18
Support C# enums

mhedgpeth:
This branch supports C# enums as a part of #146 

Since in C# the primitive-only enums support copy-only I was careful to support that to maximize performance; that performance consideration that we want the absolute best if possible, does create a lot of complexity in the generator but I think it's warranted.

I also rebased to get the benchmark working, and took a different approach than #189 - by staying at uniffi 0.31 (which is a more honest comparison) and making `uniffi-bindgen-cs` tied to [this PR](https://github.com/NordSecurity/uniffi-bindgen-cs/pull/163) for now, which is a valid PR, but hasn't received attention by the maintainer(s) in order to merge. (And this is a major reason why I am here right now doing this, I like the idea of a unified FFI approach and like what boltffi is doing for performance).

The benchmarks work, the performance is outstanding except for it's at parity with the `GenerateString64K` and `EchoString64K` cases, which are expected due to the parity in approach. Everything else is significantly better on the benchmark test.

I like the benchmark the way it is and suggest closing #189 and going with this; I'll wait for @engali94 before proceeding.


engali94:
Thanks @mhedgpeth  for doing this your approach is way better than the compat downgrade path I used in #189, I closed #189 in favor of this.

mhedgpeth:
@engali94 ready for your review

boltffi_bindgen/src/render/csharp/lower.rs
```diff
@@ -150,18 +235,51 @@ impl<'a> CSharpLowerer<'a> {
...
+    /// enums as heap objects and can't claim the same equivalence.
     fn is_blittable_record(&self, id: &RecordId) -> bool {
-        self.abi_record_for(id).is_some_and(|r| r.is_blittable)
+        if self.abi_record_for(id).is_some_and(|r| r.is_blittable) {
+            return true;
+        }
+        let Some(definition) = self.ffi.catalog.resolve_record(id) else {
+            return false;
+        };
+        definition.is_repr_c
```
engali94: Are we widening this more than intended? `supported_enums` includes data enums now, so this would mark a `#[repr(C)]` record with a data enum field as blittable on the C# side even though the IR still treats data enums as wire encoded
this looks like a false positive for the direct `P/Invoke` path

so a shape like this becomes possible:

```rust

#[data]
#[derive(Clone, Debug, PartialEq)]
pub enum Shape {
    Circle {
        radius: f64,
    },
...
}

#[data]
#[repr]
pub struct Holder {
    shape: Shape, 
}

```
the current C# check treats `shape: Shape` as if it were a blittable enum field

boltffi_bindgen/src/render/csharp/lower.rs
```diff
@@ -243,6 +375,273 @@ impl<'a> CSharpLowerer<'a> {
...
+            return None;
+        }
+        let class_name = NamingConvention::class_name(enum_def.id.as_str());
+        let methods = self.lower_enum_methods(enum_def, &class_name);
+        match &enum_def.repr {
+            EnumRepr::CStyle { tag_type, variants } => {
+                let lowered_variants = variants
+                    .iter()
+                    .enumerate()
+                    .map(|(ordinal, variant)| CSharpEnumVariant {
```
engali94, line 393: it is using the variant ordinal for the public C# enum members instead of the Rust discriminant we keep those two concepts separate where the public enum value matches the Rust discriminant while the wire codec uses ordinal tags
here we are using the ordinal for the public enum too so this happens to work for the current `0,1,2...` examples but it will break as soon as we have a C style enum with gapped discriminants because direct enum params and returns depend on the numeric value matching Rust exactly

we should probably extend the demo with one of those cases to get it covered end to end.

mhedgpeth: Thanks for the careful review. I added test cases for each scenario, wired them through every backend harness (where the non-C# ones passed as-is), and fixed C# to match.

Two follow-ups I'd like to tackle after this merges:

1. **`#[repr(C)]` struct with a C-style enum field → direct P/Invoke.** `TaskHeader` in this PR lays out identically in Rust and in a correctly-annotated C# / Swift / Kotlin / Java struct, so it should be able to skip wire encoding. The blocker is in `boltffi_macros`: the proc-macro's blittability check only admits fields whose *type syntax* matches a known primitive name. A field typed `Priority` (a C-style enum with `#[repr(i32)]`) bounces out even though its layout is identical. Fixing this needs a macro-level opt-in signal (attribute, registry, etc.) plus matching widening in each backend's lowerer.

2. **Width-correct wire format for C-style enums.** Every backend currently wire-encodes a C-style enum's ordinal tag as a 4-byte `i32`, regardless of `#[repr(iN)]`. `LogLevel` is `#[repr(u8)]` — one byte on the wire would suffice. Worth reclaiming in batched/`Vec<T>` workloads. Requires coordinated changes to `boltffi_macros`, the IR lowerer, and every backend's enum template.

Both are cross-language changes, so they deserve their own PRs with their own before/after byte-level coverage. Would you like me to make issues for each since they kind of go out of scope of the C# change? And do you think these are appropriate for me to try to tackle?

engali94: Thanks! the two follow ups make sense to me and I agree they should stay out of scope for this PR they deserve their own issues and their own coverage.

and for the `#[repr(u8)] #[repr(u16)]` side we have had some discussion around that before and I do think we should bring it back up and support it properly across all widths

engali94 (approved):

PR #200, opened 2026-04-20
Support python blittable records

osa8bit:


boltffi_bindgen/src/render/python/templates/render_python/init_py.txt
```diff
@@ -23,6 +27,35 @@ def _shared_library_filename() -> str:
 
 _native._initialize_loader(str(Path(__file__).resolve().with_name(_shared_library_filename())))
 {% endif %}
+{% if module.uses_records() %}
+
+{% for record in module.records %}
+@dataclass(frozen=True, slots=True)
```
engali94, line 33: it turned out while testing that slots=True requires pythn > 3.10
do we pin the python version anywhere in the generated package?

engali94 (approved):

boltffi_bindgen/src/render/python/lower/records.rs
```diff
@@ -0,0 +1,289 @@
...
+            &methods,
+        )?;
+
+        Ok(Some(
+            PythonRecord::new(type_ref, fields, constructors, methods)
+                .expect("direct python record should always have at least one field"),
+        ))
+    }
+
+    fn lower_record_field(field: &crate::ir::definitions::FieldDef) -> PythonRecordField {
```
engali94: nit: let us move this import to the top

engali94 (approved):

boltffi_bindgen/src/render/python/templates/render_python/init_py.txt
osa8bit: yes there will be version gating starting with classes impl

PR #205, opened 2026-04-22
Support C# Option<T>

mhedgpeth:
This PR supports Option<T> as a part of #146

Wire format is unchanged; Option<T> is just a one byte (1 = some, 0 = none) with an encoded payload.
`Option<Option<T>>` rejected because `string??` isn't valid in C#.
Added tests for coverage, TypeScript failed and so I will make an issue for that.

I'm allowing some things with clippy that I'll come back to because I want to wait for the design to finalize before I make sure it's cleaned up, and because I see some of those same things with the java backend.


boltffi_bindgen/src/render/csharp/emit.rs
```diff
@@ -350,13 +374,37 @@ fn csharp_type_for_inner(ty: &TypeExpr) -> String {
...
 }
 
+/// C# type literal for the unwrapped payload of a `ReadOp::Option`.
+/// Used by the Option read emit to stamp a `(T?)null` cast on the null
+/// branch of the ternary — without it, the compiler infers the
+/// conditional as `int` (value types) or bare `null` (references),
+/// both of which fail to round-trip back into `int?` / `string?`.
+/// Derived from the Option's own inner `ReadSeq` so emit does not
+/// need the original `TypeExpr` threaded alongside the seq.
+fn csharp_type_for_read_seq_inner(seq: &ReadSeq) -> String {
```
engali94: this seems to be re-implementing the same functionality in `csharp_type_for_inner ` and `CSharpType::qualify_if_shadowed()` we need to find a way to collapse them into some kind of strongly typed helper to prevent drift

boltffi_bindgen/src/render/csharp/emit.rs
```diff
@@ -420,6 +480,15 @@ fn emit_reader_read_with_context(
             let class_name = NamingConvention::class_name(id.as_str());
             format!("{}.Decode(reader)", qualify_if_shadowed(&class_name, scope))
         }
+        ReadOp::Option { some, .. } => {
+            let inner = emit_reader_read_with_context(some, scope, ctx);
+            let inner_type = csharp_type_for_read_seq_inner(some);
+            format!(
+                "reader.ReadU8() == 0 ? ({inner_type}?)null : {inner}",
```
engali94: i think this still misses the shadowing case inside data enum bodies where if a variant name collides with a module level type the `Option<T>` decode here still casts to `T?` so inside the enum body C# binds that to the nested type instead of the outer one

boltffi_bindgen/src/render/csharp/lower.rs
```diff
@@ -273,6 +280,17 @@ impl<'a> CSharpLowerer<'a> {
...
             }
+            ReturnDef::Value(TypeExpr::Option(_)) => {
+                // Fully pre-render the decode expression against a
+                // reader named `reader`. The ABI's decode_ops walks
+                // every inner shape — primitive, string, record, enum,
+                // vec — so one helper covers the whole matrix.
+                let decode_seq = decode_ops
+                    .expect("Option return must carry decode_ops")
+                    .clone();
+                let decode_expr = emit::emit_reader_read(&decode_seq, None);
```
engali94: I think the problem here is that the `Option<T>` return decode gets rendered too early here with `emit_reader_read(..., None)` it works for free functions but if it is emitted inside a data enum body, we willl the shadowing scope so the generated (`Point?)null` cast will bind to the nested type instead of the outer one

mhedgpeth:
I'm going to centralize type-based emission around the `CSharpType` type (and remove the two drifting helper methods), and I'm going to add a couple of additional demo tests to expose the issue in the third comment, this will be over several commits.

boltffi_bindgen/src/render/csharp/emit.rs
mhedgpeth: Good catch, I consolidated to `CSharpType`

boltffi_bindgen/src/render/csharp/emit.rs
mhedgpeth: Latest commit fixes this

boltffi_bindgen/src/render/csharp/lower.rs
mhedgpeth: I added a test case for this, corrected it

mhedgpeth:
I consolidated to the `CSharpType` and the build is now working, but I'm going to do a broader review of the design because I want to be clearer about the roles of each of the modules, but also I want to be consistent with the other backends, before merging / another review I'll decide on my approach and put it here / update the module documentation.

mhedgpeth:
@engali94 OK thanks for the comments, I worked on getting the design to fall into a pit of success easier, and got clearer about what I want to do here.

The fundamental issue here is making sure the roles are correct:

I receive an IR from the framework, no arguments there.
I use `lower` to make _my own_ IR (IIR?) at the csharp level, right now this centers around `CSharpType`. The improvements I made migrated disparate naming methods into the central type.

Then `emit` should use the `CSharpType` as a view model of sorts to output via the template, to keep the template from having a ton of conditional logic in it.

I don't like it that `lower` depends on `emit` to determine names in advance. In my mind, `lower` creates the _language-specific_ Intermediate Representation (IR between the framework IR and the emission), and that's it. Then the `emit` takes the already calculated `CSharpType` (and associated friends), and emits them without a care in the world.

When I got to that last idea, I thought that it was out of scope of the PR, so I wrapped it up, included tests to ensure coverage, and am calling this ready to go. The next PR will be that refactoring, but I wanted to write it out to make sure we're aligned, since the other areas don't follow that strict pattern. Let me know what you think, I'll wait to merge before hearing from you.

engali94 (approved):
Thanks, this direction makes sense to me. I agree with the layer split you described, That is the direction I wanted us to move toward too, because the earlier bugs came from type/name decisions being split across multiple places.

I’m happy for you to take that followup in the direction you think is cleanest, as long as we keep the layer boundaries explicit. If the C# backend lands on a better pattern, we can use that as a reference for improving the other backends over time.

Separately I am also working on some architecture sketches for making the backend layer more IR and compiler driven, so backends have less room to drift and unsupported cases are caught more systematically. I’ll open a discussion for that once the shape is clearer. That is much broader than this PR, but it lines up with the same concern

mhedgpeth:
@engali94 fantastic we're on the same page. I'll be careful to keep it in the `csharp` section and hopefully will spring up some good ideas that might be pushed further up when we consider consolidation.

PR #207, opened 2026-04-22
Speed up python enum boxing and decoding

osa8bit:
| Benchmark | BoltFFI before ns/op | BoltFFI after ns/op | BoltFFI improvement | UniFFI after ns/op | BoltFFI after vs UniFFI |
| --- | ---: | ---: | ---: | ---: | ---: |
| `simple_enum` | 258.46 | 87.85 | 2.94x faster | 11825.71 | 134.62x faster |
| `echo_direction` | 222.87 | 54.52 | 4.09x faster | 6973.22 | 127.91x faster |
| `generate_directions_1k` | 205447.67 | 3367.64 | 61.01x faster | 280191.71 | 83.20x faster |
| `generate_directions_10k` | 1769052.31 | 32200.53 | 54.94x faster | 2791636.72 | 86.70x faster |
| `count_north_1k` | 6610.76 | 6667.83 | essentially unchanged | 1681916.86 | 252.24x faster |
| `count_north_10k` | 64517.13 | 65395.58 | essentially unchanged | 16668805.21 | 254.89x faster |


boltffi_bindgen/src/render/python/templates/render_python/native_module.c.txt
```diff
@@ -538,6 +554,93 @@ static int {{ primitive.vector_parser_name() }}(PyObject *value, boltffi_python_
...
+    const char *enum_name,
+    const char *const *member_names
+) {
+    PyObject **loaded_members = NULL;
+    Py_ssize_t member_index = 0;
+    if (!PyType_Check(type_object)) {
+        PyErr_Format(PyExc_TypeError, "expected type for %s", enum_name);
+        return 0;
+    }
+    loaded_members = PyMem_Calloc((size_t)registration->member_count, sizeof(PyObject *));
```
engali94, line 675: I think registration needs to validate the cached enum members not just look them up by name becuase a user could register an IntEnum that has the right names but wrong values like NORTH = 10 instead of NORTH = 0 thus the native returns would now silently box to that wrong member. 
The old `type_object(native_value)` was rejecting that

boltffi_bindgen/src/render/python/templates/render_python/native_module.c.txt
```diff
@@ -538,6 +554,93 @@ static int {{ primitive.vector_parser_name() }}(PyObject *value, boltffi_python_
...
+    const char *enum_name
+) {
+    if (registration->type_object != NULL) {
+        return 1;
+    }
+    PyErr_Format(PyExc_ImportError, "native type %s is not registered", enum_name);
+    return 0;
+}
+
+static int boltffi_python_expect_enum_instance(
```
engali94, line 701: This is starting to duplicate the registered type validation path we already have for records just below the enum path needs the member cache but the basic `type_object` and instance checks feel shared. I would rather avoid two parallel registration flows in the template if we can

boltffi_bindgen/src/render/python/emit/emitter.rs
```diff
@@ -564,11 +564,23 @@ mod tests {
...
         assert!(native_source.contains("static PyObject *boltffi_python_box_status_from_wire_tag"));
         assert!(native_source.contains("static int boltffi_python_parse_vec_status"));
         assert!(native_source.contains("static PyObject *boltffi_python_decode_owned_vec_status"));
+        assert!(native_source.contains("typedef struct boltffi_python_c_style_enum_registration"));
+        assert!(
+            native_source
+                .contains("static PyObject *boltffi_python_direction_members_by_wire_tag[2]")
+        );
+        assert!(
+            native_source.contains("static const char *boltffi_python_direction_member_names[2]")
```
engali94, line 577: this doesn't test the new behavior it only check if the C symbols exist, the important behavior is `wire_tag -> cached member` instead of `type_object(native_value)`

engali94 (approved):

PR #212, opened 2026-04-23
Fix async exports with non-primitive data params

### LINKED ISSUE #208: async #[export] method on &self fails to compile with #[data] param containing non-primitive fields
### Repro

```rust
use boltffi::{data, export};

#[data]
pub struct Input {
    pub s: String,
}

pub struct Service;

#[export]
impl Service {
    // compiles
    pub fn sync_method(&self, input: Input) -> bool {
        true
    }

    // compiles
    pub async fn static_method(input: Input) -> bool {
        true
    }

    // fails
    pub async fn async_method(&self, input: Input) -> bool {
        true
    }
}
```

### Error

```
error[E0308]: mismatched types
  --> src/lib.rs:10:1
   |
10 | #[export]
   | ^^^^^^^^^ expected `()`, found `*const _`
   |
   = note: expected unit type `()`
            found raw pointer `*const _`
   = note: this error originates in the attribute macro `export` (in Nightly builds, run with -Z macro-backtrace for more info)
help: you might have meant to return this value
   |
10 | return #[export];
   | ++++++          +

error[E0308]: mismatched types
  --> src/lib.rs:10:1
   |
10 | #[export]
   | ^^^^^^^^^ expected `Input`, found `*const _`
   |
   = note:   expected struct `Input`
           found raw pointer `*const _`
   = note: this error originates in the attribute macro `export` (in Nightly builds, run with -Z macro-backtrace for more info)
```

### Possible regression

Compiles on 0.2.0. Fails on 0.21.0, 0.22.0, 0.22.1, 0.23.0, 0.23.1, 0.24.0. Tested for aarch64-apple-darwin and wasm32-unknown-unknown.

### Environment

boltffi: 0.24.0
rustc: 1.95.0
os: macOS

engali94:
fixes #208

### DIFF
--- a/boltffi_bindgen/src/render/jni/templates/jni_glue.txt
+++ b/boltffi_bindgen/src/render/jni/templates/jni_glue.txt
@@ -39,6 +39,16 @@ static inline void boltffi_throw_runtime(JNIEnv* env, const char* message) {
     (*env)->DeleteLocalRef(env, exception_class);
 }
 
+static inline void boltffi_throw_status(JNIEnv* env, FfiStatus status, const char* fallback_message) {
+    if (status.code == 3) {
+        boltffi_throw_illegal_argument(env, "invalid argument");
+    } else if (status.code == 4) {
+        boltffi_throw_runtime(env, "operation cancelled");
+    } else {
+        boltffi_throw_runtime(env, fallback_message);
+    }
+}
+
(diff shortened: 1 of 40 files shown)


mhedgpeth (approved):
This is a bit ahead of where I am but I'm happy to read it through and continue to learn more about the crate. It looks good.

PR #216, opened 2026-04-24
feat(kotlin): add configurable desktop native loader mode

Aunali321:
Adds a Kotlin generator option for desktop JVM native loading policy.

Motivation: the current Kotlin bindings always enable BoltFFI’s desktop loader on non-Android JVMs. That works for bundled desktop artifacts, but it does not fit hosts that already manage native library loading before the generated bindings are initialized. This change makes the behavior explicit and configurable with `targets.android.kotlin.desktop_loader`, while preserving existing behavior as the default.

Supported modes:
- `bundled` (default): current BoltFFI behavior
- `system`: use `System.loadLibrary` on desktop JVMs
- `none`: skip desktop loading and assume the host already loaded the native library


engali94 (approved):
LGTM

PR #246, opened 2026-04-25
Refactor C# Structure

mhedgpeth:
This refactors the structure within the `render/csharp` without touching the IR, with the following principles:

1. Replace `String` with a [new type](https://effective-rust.com/newtype.html) - in the previous design `String` meant a lot of different things which led to confusion.
2. Created an `ast` package to represent C# syntax primitives, without building the entire AST. Most of these are the new types over `String` but make it much easier to reason about what's happening
3. This created a clearer `plan` module, with `*Plan` structures that are centered around what we are rendering, as a ViewModel of the template.
4. This created a clearer `templates` implementation where the templates take on most of the generation while properly deferring the variable and decision-tree parts to the underlying `*Plan` objects
5. Refactored `lower` to conform to the recent python refactoring, splitting the implementation up by function.
6. Reviewed documentation to be less verbose, but still present


engali94:
I took a quick pass on it, looks amazing @mhedgpeth I'll do proper review later today

boltffi_bindgen/src/render/csharp/ast/code.rs
```diff
@@ -0,0 +1,727 @@
...
+/// A C# binary operator token.
+///
+/// Examples:
+/// ```csharp
+/// a == b
+/// a + b
+/// a * b
+/// ```
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub(crate) enum CSharpBinaryOp {
```
engali94, line 88: this is very nice, probably making it shared makes more sense

boltffi_bindgen/src/render/csharp/ast/code.rs
```diff
@@ -0,0 +1,727 @@
...
+/// ```csharp
+/// this.X
+/// reader.ReadF64()
+/// (int?)null
+/// reader.ReadU8() == 0 ? (int?)null : reader.ReadI32()
+/// item => item.Decode()
+/// this.Name is { } opt0
+/// ```
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub(crate) enum CSharpExpression {
```
engali94, line 117: The DSL combined with `CSharpStatement` is a really nice pattern Im thinking of generalizing it and reusing it for other backends.

engali94 (approved):
really nice cleanup! Feel free to ignore the comments, I just added them for reference.

PR #249, opened 2026-04-26
fix(jni): don't look up boltffiFutureContinuationCallback for sync-only callback traits

### LINKED ISSUE #250: Android crash on load: JNI_OnLoad fails when crate has only sync callback traits
## Bug

A crate with only sync `#[export]` callback traits and no async functions crashes on Android when the native library is loaded.

```rust
#[export]
pub trait EventHandler: Send + Sync {
    fn on_event(&self, event: MyEvent);
}

#[export]
impl Engine {
    pub fn new(handler: Box<dyn EventHandler>) -> Self { ... }
}
```

The app crashes immediately — before any user code runs.

## Cause

`JNI_OnLoad` is emitted whenever `has_async` is true, which includes crates with callback traits. Inside `JNI_OnLoad`, `boltffiFutureContinuationCallback` is always looked up — but the Kotlin codegen only emits that method when `has_async_runtime` is true (async functions/methods/streams). For a sync-only callback trait `has_async_runtime` is false, so the method doesn't exist in Kotlin and `JNI_OnLoad` returns `JNI_ERR`.

## Workaround

Add a dummy async export to force `has_async_runtime = true`:

```rust
#[export]
pub async fn noop() {}
```

Fix tracked in #249.

Rajaniraiyn:
Fixes #250.

`has_async` is true whenever callback traits exist, so `JNI_OnLoad` was unconditionally looking up `boltffiFutureContinuationCallback`. The Kotlin codegen only emits that method when `has_async_runtime` is true (async functions, async methods, or streams) — not for sync-only traits.

Result: any crate with only sync callback traits crashes on Android at library load (`JNI_OnLoad` returns `JNI_ERR`).

**Fix:** split `has_async_runtime` out of `has_async` in the JNI lowerer, mirroring the condition already used on the Kotlin side, and gate the continuation callback lookup on it.


Rajaniraiyn:
https://github.com/boltffi/boltffi/pull/249/commits/f558bbca0ce25b4617b6570bb1eb85b16f2ce926 fixes failed https://github.com/boltffi/boltffi/actions/runs/24955056226/job/73073142730

engali94 (approved):
Looks goo, thank you!

PR #251, opened 2026-04-26
feat(dart): scaffold support for dart records

nxmxsxs:
- Add typedefs for custom types
- scafford rendering for dart records
- refactor WireReader/WireWriter: Read/Write directly from/to memory without going through dart's ByteData


mhedgpeth:
Build fails, assigning to @nxmxsxs

engali94 (approved):

PR #252, opened 2026-04-26
Fix Android library loading bug

### LINKED ISSUE #247: Version 0.24.1 causes failed to load library on android
Android load library on version 0.24.1 causes exception on library load.

### Symptom
When the first method is called on Kotlin side, it will try to load .so library, but it will crash the app since the ``fallbackLibrary`` property on .kt generated file points to  ``sample_library`` instead of ``sample-library`` that was generated on version 0.23.1 or less. Old versions it respects the value set on cargo.toml/boltffi.toml for package-name

### Workaround
Simple renaming the value on .kt generated file eg:

boltffi.toml \ cargo.toml
```
[package]
name = "sample-library"
```

Current Output
```
val preferredLibrary = "sample_library_jni"
val fallbackLibrary = "sample_library"
```

Expected Output
```
val preferredLibrary = "sample-library-jni"
val fallbackLibrary = "sample-library"
```

ahmedre:
Fixes #247.

### DIFF
--- a/boltffi_bindgen/src/render/kotlin/lower.rs
+++ b/boltffi_bindgen/src/render/kotlin/lower.rs
@@ -29,6 +29,8 @@ use crate::ir::plan::{AbiType, ScalarOrigin, SpanContent, Transport};
 use crate::ir::types::{PrimitiveType, TypeExpr};
 use crate::render::kotlin::emit;
 use crate::render::kotlin::plan::*;
+#[cfg(test)]
+use crate::render::kotlin::templates::KotlinEmitter;
 use crate::render::kotlin::templates::{AsyncMethodTemplate, WireMethodTemplate};
 use crate::render::kotlin::{
     FactoryStyle, KotlinApiStyle as KotlinInputApiStyle, KotlinDesktopLoader, KotlinOptions,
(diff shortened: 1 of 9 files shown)


boltffi_bindgen/src/render/kotlin/lower.rs
```diff
@@ -4989,6 +4995,64 @@ mod tests {
...
+        let contract = FfiContract {
+            package: PackageInfo {
+                name: "sample-library".to_string(),
+                version: None,
+            },
+            catalog: TypeCatalog::default(),
+            functions: vec![],
+        };
+        let module = lower_module(&contract);
+        let rendered = crate::render::kotlin::templates::KotlinEmitter::emit(&module);
```
engali94: nit: let us move this to the top f the module

boltffi_bindgen/src/render/kotlin/lower.rs
```diff
@@ -4989,6 +4995,64 @@ mod tests {
...
+            &abi,
+            "com.example.demo".to_string(),
+            "demo".to_string(),
+            KotlinOptions {
+                library_name: Some(naming::load_library_name("configured-library")),
+                ..KotlinOptions::default()
+            },
+        )
+        .lower();
+        let rendered = crate::render::kotlin::templates::KotlinEmitter::emit(&module);
```
engali94: similar here

engali94 (approved):

PR #255, opened 2026-04-27
Support C# classes

mhedgpeth:
This PR supports C# classes as a part of #146 

It uses an `IntPtr` to track the value of a struct in memory. Demo tests pass, but there are a couple of cases that I couldn't cover that seem to be broader gaps, I'll file an issue for those (#256).

I'm getting to the phase of my boltffi implementation journey where I keep asking myself [this question](https://www.youtube.com/watch?v=ti9bzNWWTDY) of the user, but I'll keep trucking along, it's fun!


boltffi_bindgen/src/render/csharp/plan/module.rs
```diff
@@ -38,22 +42,39 @@ impl CSharpModulePlan {
...
 
-    /// Whether any function takes a wire-encoded record param. Blittable
-    /// record params pass through the CLR as direct struct values and do
-    /// not contribute here.
+    /// Whether any function or class constructor takes a wire-encoded
+    /// record param. Blittable record params pass through the CLR as
+    /// direct struct values and do not contribute here.
     fn has_wire_params(&self) -> bool {
         self.functions.iter().any(|f| !f.wire_writers.is_empty())
+            || self.classes.iter().any(|c| {
```
engali94: if the rust crate doesnt export free functions but only classes with methods it fails because this only folds class constructors into the helper emission predicates but class methods can also need these helpers as well any class class method with encoded params needs `WireWriter` and others

boltffi_bindgen/src/render/csharp/snapshots/boltffi_bindgen__render__csharp__templates__tests__snapshot_class_counter_with_methods.snap
```diff
@@ -0,0 +1,55 @@
...
+        {
+            _handle = handle;
+        }
+
+        public int Get()
+        {
+            return NativeMethods.CounterGet(_handle);
+        }
+
+        public void Increment()
```
engali94, line 31: if we call Increment() after Dispose() `_handle` is `IntPtr.Zero`  then it will pass null into Rust and crashes instead of throwing some exception worth a disposed check on the wrappers?

boltffi_bindgen/src/render/csharp/lower/classes.rs
```diff
@@ -0,0 +1,209 @@
...
+        let surface_name = match &kind {
+            CSharpConstructorKind::Primary => CSharpMethodName::new("New"),
+            CSharpConstructorKind::StaticFactory { name } => name.clone(),
+        };
+        let native_method_name = CSharpMethodName::native_for_owner(class_name, &surface_name);
+        let helper_method_name = match &kind {
+            CSharpConstructorKind::Primary => {
+                CSharpMethodName::new(format!("{class_name}NewHandle"))
+            }
+            CSharpConstructorKind::StaticFactory { .. } => CSharpMethodName::new(""),
```
engali94: Static factories don't have a helper method so we end up writing `CSharpMethodName::new("") `as a sentinel, the template happens not to read it as well
could we model this in the type instead? either store the helper name only on the `Primary` variant or make it `Option<CSharpMethodName>`?

engali94 (approved):

boltffi_bindgen/src/render/csharp/snapshots/boltffi_bindgen__render__csharp__templates__tests__snapshot_class_counter_with_methods.snap
mhedgpeth: I'm going to follow the dotnet convention and throw a `ObjectDisposedException` if something is called after `Dispose` is called

boltffi_bindgen/src/render/csharp/lower/classes.rs
mhedgpeth: part of: [C# - move helper_method_name into Primary constructor variant](https://github.com/boltffi/boltffi/pull/255/commits/4495e65e471f2b740e3951df28e54bbc5b1be394)

boltffi_bindgen/src/render/csharp/plan/module.rs
mhedgpeth: fixed in: 
C# - fold class methods into module-level helper-emission predicates

boltffi_bindgen/src/render/csharp/snapshots/boltffi_bindgen__render__csharp__templates__tests__snapshot_class_counter_with_methods.snap
mhedgpeth: [C# - throw ObjectDisposedException from class methods after Dispose](https://github.com/boltffi/boltffi/pull/255/commits/f5283c6c5b096b607d4d7231894dde3d70d275c5)

PR #259, opened 2026-04-27
Add support for python records

osa8bit:


engali94:
@osa8bit holding off on this for now, special casing and arch smells in here. can we sync? 
also thinking python as the first target to experiment with the new AST/Binding DSL

engali94:
must be reworked to follow the IR based approach

PR #260, opened 2026-04-27
Propose binding IR for the backend layer

engali94:
This PR opens an ADR proposing a redesign of the bindgen backend layer. The current architecture lets nine backends re-derive the same ABI decisions independently, producing drift the compiler cannot catch. The ADR argues for a sealed binding IR that turns those decisions into the shape of the data backends receive, so disagreement becomes a build error rather than a runtime mismatch. Would love feedback on the proposal.

### DIFF
--- a/adrs/0001-adr-binding-ir-architecture.md
+++ b/adrs/0001-adr-binding-ir-architecture.md
@@ -1,0 +1,1348 @@
+# ADR: Binding IR. One Source of Truth Between Contracts and Backends
+
+## Status
+Proposed
+
+## Authors
+- [Ali Hilal](githhub.com/engali94)
+
+## Date
+2026-04-27
+
+---
+
+## 1. Background and Problem
+
+### 1.1 The pipeline today
+
+```
+   user crate (Rust)
+        |
+        |  #[data], #[export], #[data(impl)] proc macros
+        v
+   boltffi_macros            (emits cfg-gated contract items)
+        |
+        |  emits per-target glue + parsed module info
+        v
+   boltffi_bindgen
+        |   builds:
+        |     FfiContract   (semantic side, in src/ir/contract.rs)
+        |     AbiContract   (boundary side, in src/ir/abi.rs)
+        v
+   render/<lang>/             (one folder per target language)
+        lower.rs   ->   plan.rs   ->   emit.rs / templates.rs
+        |
+        v
+   generated source files (Swift, Kotlin, Java, JNI, C#, Dart,
+                           TypeScript, Python, C)
+```
+
+Three stages produce one stream of files. The third stage exists nine times, once per target language, and is the part this ADR is about.
+
+### 1.2 What the user writes
+
+A real example from `examples/demo/src/records/blittable.rs`:
+
+```rust
+use boltffi::*;
+
+/// A 2D point with double-precision coordinates.
+#[data]
+#[derive(Clone, Copy, Debug, PartialEq, Default)]
+pub struct Point {
+    pub x: f64,
+    pub y: f64,
+}
+
+#[data(impl)]
+impl Point {
+    pub fn new(x: f64, y: f64) -> Self          { Point { x, y } }
+    pub fn origin()              -> Self          { Point { x: 0.0, y: 0.0 } }
+    pub fn distance(&self)       -> f64           { (self.x * self.x + self.y * self.y).sqrt() }
+    pub fn try_unit(x: f64, y: f64) -> Result<Self, String> { /* ... */ }
+}
+
+#[export]
+pub fn make_point(x: f64, y: f64) -> Point { Point { x, y } }
+
+#[export]
+pub fn add_points(a: Point, b: Point) -> Point {
+    Point { x: a.x + b.x, y: a.y + b.y }
+}
+```
+
+The user expects this to produce a `Point` type, a `make_point` free function, and an `add_points` free function in every target language, all wired to the same Rust crate, all interoperating correctly.
+
+### 1.3 What the bindgen produces from it
+
+From `boltffi_bindgen/src/ir/contract.rs`, the semantic side:
+
+```rust
+pub struct FfiContract {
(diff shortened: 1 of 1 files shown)


mhedgpeth:
It's great that you're thinking on this level, thanks for the proposal. My thoughts:

1. I suggest there be a strict representation of what is in rust in an AST, and that's that. Nothing more. So the rest of the program can ask "what did the user in rust say" - and can get an answer.
2. In my mind shoring up the tests would be a valuable thing to do, the tests are your compiler, because you're dealing with a cross-language problem. I was thinking of cataloging in demo (rust) each test case, you could create a central tests.toml that outputs what tests are expected in each backend, with an id and then refer to that id where it's implemented, then have CI ensure that all ids are covered for all (supported) render languages. This to me would be more of a "compiler" safety net than anything, IMO. The tests in this repo are essential and powerful.
3. Each emission needs to be able to (A) understand bindings to other things, and optionally put a namespace on it, (B) make an independent decision about things, because languages are different, for example I don't believe there is a single answer for whether a record is blittable, in the demo, `Point { x: f64, y: f64 }` is passed by value at the rust ABI. C# can render it with `[StructLayout(Sequential)]` and Swift with `@frozen` but Kotlin and Java go through JNI, which can't pass structs by value at all; they have to marshal field by field. Same record, different answers.
4. Ordering of emission also differs by language. C needs types forward-declared before any function uses them; Kotlin can emit in any order. Swift wants extensions grouped at the end of the file. I'm not sure you can centralize it.

I suggest a layered strategy:

A - strict AST for what's in rust, the `FfiContract` claims to be this but the `is_blittable` fn in `RecordDef` drifts it from "here's what the user told me" - I think there is an abstraction leak here that could be shored up. In short, `ir/definitions.rs` should not use `boltffi_ffi_rules` or `ir/abi` - let that abstraction sit alone, and then add to it, even if by extension traits. 
B - shared FFI primitives: codecs, name canonicalization, symbol manifests, capability flags. Contains the descriptive material every backend can pull from, on an as needed basis.
C - language-specific render/emission
D - tests

The goal is obviously to minimize C while preserving essential flexbility. So push those out. My priorities would be:

1. Standardize D with a `tests.toml` that declares what demo elements are required and what they test, and have languages declare in their own tests (through comments) their adherence to that. This gives you safety to make changes to C. It's a problem that there are few standards about coverage, and I keep finding gaps, which tells me there is work to do here.
2. Ensure A is in place which will decouple centralized FFI rendering decisions from the declaration by the user
3. Look for decisions among the renders that deviate, and make the API for B easier for those, one at a time

engali94:
Thanks @mhedgpeth , this makes sense. I think there are two related points here.

For point 1. agreed on the goal. We want `FfiContract` to be a strict AST that just describes what the user wrote in Rust, no decisions baked in. The whole reason `binding/` exists is to make that cleanup possible.

Today `FfiContract `feeds nine backends directly. That's why `is_blittable` and the rest of the leakage is still there, stripping it now means refactoring every backend at once, which is the kind of change that doesn't land. `binding/ ` moves the decision layer up and out, so the resolver becomes the only thing reading `FfiContract. After that, turning the AST into pure description is an isolated change. That's the next step, and it's the point of doing this one first.ry shape did resolution choose?”

For the testing point: agreed on this one too. If I'm reading you right, what you're proposing is roughly:

```toml
  [[case]]
  id = "record.direct.point"
  source = "examples/demo/src/records/blittable.rs"
  requires = ["swift", "kotlin", "java", "python", "csharp"]
```
so each backend writes its test in its own idiomatic style. The catalog just enforces coverage.
That's the right move and lands quickly. 

 Worth mentioning the heavier alternative we'd looked at on a separate track with @osa8bit, which never landed, a TOML that holds the Rust source plus inputs and expected outputs, with per-language tests auto-generated and run at runtime. Bigger payoff in theory (single source of truth, catches runtime drift between backends, scales linearly with cases instead of cases times backends). The idea was Roughly:
```
test_cases/
  primitives.toml
  callbacks.toml
  async.toml
  structs.toml
  enums.toml
```

where each manifest declares:
- Rust code to export
- Test inputs
- Expected outputs

### 2. Manifest Schema

```toml
[meta]
name = "primitives"
description = "Test primitive type passing"

[[test]]
name = "add_integers"
rust = """
#[export]
pub fn add(a: i32, b: i32) -> i32 { a + b }
"""

[[test.case]]
inputs = { a = 1, b = 2 }
expected = 3

[[test.case]]
inputs = { a = -1, b = 1 }
expected = 0

[[test.case]]
inputs = { a = 2147483647, b = 0 }
expected = 2147483647
```
then we call `boltffi test`

Then boltffi test runs the suite across every backend. Side benefit of it was that BoltFFI users get the same harness for testing their own FFI boundaries.
but also this approach had risks ex. a bug in the generator gives false positives in every backend at once. The generator becomes a thing to maintain, and source code is in TOML raw strings is a burden. I'd keep that as an exploration path for later if needed. 

Either way, I think this is big enough to be separate from the current ADR.

mhedgpeth:
Regarding the refactoring - yes you're right but why not start there, choke it off one at a time? I don't mind getting PRs that do this and having to reconcile.

Then with a clear separation you'll be in better shape to understand the next step.

Regarding testing I created #264 and will propose something, agree that it shouldn't go here, I think the reason I mentioned it was because it was in the context of "what can reduce duplication/increase streamlining of the backends"

engali94:
> Regarding the refactoring - yes you're right but why not start there, choke it off one at a time? I don't mind getting PRs that do this and having to reconcile.
> 
> Then with a clear separation you'll be in better shape to understand the next step.
> 
> Regarding testing I created #264 and will propose something, agree that it shouldn't go here, I think the reason I mentioned it was because it was in the context of "what can reduce duplication/increase streamlining of the backends"

Makes sense. I'll extend the ADR to cover the AST piece as well, and start opening small PRs to chip away at the leaks one at a time thanks, this was really helpful.

mhedgpeth (approved):

PR #272, opened 2026-04-29
Introduce a dedicated AST crate

engali94:
Introduces `boltffi_ast` as a leaf crate holding the pure AST types describing what the user wrote in Rust,  the crate models exported Rust declarations, type expressions, names, source metadata, docs, attributes, and IDs without wiring it into the existing bindgen pipeline yet.


mhedgpeth (approved):
This is great!

osa8bit (approved):
nice!

PR #273, opened 2026-04-29
feat: C# - support record instance methods and static factories

mhedgpeth:
This PR adds instance methods and static factories for C# as a part of #146. There was already some of this here, I added the rest, found a gap in the demo: the demo had record-method coverage, but not record methods with parameters that force the renderers to emit extra setup code/imports. So I added that, and support in the other languages that could support it.


boltffi_bindgen/src/render/csharp/templates/render_csharp/record_method.txt
```diff
@@ -0,0 +1,75 @@
+
+{%- if method.receiver.is_static() %}
+{{ self::summary_doc_block(method.summary_doc, "        ") }}        public static {{ method.return_type }} {{ method.name }}({% for p in method.params %}{{ p.csharp_type }} {{ p.name }}{% if !loop.last %}, {% endif %}{% endfor %})
+{%- else %}
+{{ self::summary_doc_block(method.summary_doc, "        ") }}        public {{ method.return_type }} {{ method.name }}({% for p in method.params %}{{ p.csharp_type }} {{ p.name }}{% if !loop.last %}, {% endif %}{% endfor %})
+{%- endif %}
+        {
+{%- for param in method.params %}
```
engali94, line 4: seems it is adding a duplicate copy of the method body logic better if we can centerlize them

engali94 (approved):

PR #274, opened 2026-05-01
C# Errors

mhedgpeth:
This PR delivers errors from `Result<T, E>` as a part of #146 

I extended the demo tests to return a primitive error, and that works on every language except for Swift.


boltffi_bindgen/src/render/csharp/lower/functions.rs
```diff
@@ -205,3 +209,324 @@ fn vec_element_read_seq(decode_ops: Option<&ReadSeq>) -> Option<ReadSeq> {
         _ => None,
     }
 }
+
+impl<'a> CSharpLowerer<'a> {
+    /// Builds the [`CSharpReturnKind::WireDecodeResult`] for a
+    /// `Result<Ok, Err>` return: pre-renders the Ok decode against a
+    /// `reader` local, plus the throw expression that constructs the
+    /// generated exception type from the wire-decoded Err payload.
+    pub(super) fn result_return_kind(
```
engali94: probably we need to move this to something like lowe/result.rs  it drives method lowering for classes, records, and enums too

boltffi_bindgen/src/render/csharp/lower/functions.rs
```diff
@@ -205,3 +209,324 @@ fn vec_element_read_seq(decode_ops: Option<&ReadSeq>) -> Option<ReadSeq> {
...
+            }
+            _ => CSharpClassName::new("BoltException"),
+        }
+    }
+
+    /// `BoltException` always takes a `string` message, so non-string
+    /// Err payloads going through the generic path need `.ToString()`.
+    /// Typed `<Name>Exception` constructors take the error value
+    /// directly and so never need to stringify it.
+    fn result_err_needs_to_string(&self, err_ty: &TypeExpr) -> bool {
```
engali94: these two methods `result_err_exception_class` and `result_err_needs_to_string` always have to agree I guess it is better to model them with something like this:
```
enum ResultErrException { Generic, Typed(CSharpClassName) }
```

engali94 (approved):

Issue #275, opened 2026-05-05
Support async #[data(impl)] methods on records and data enums


mhedgpeth:
## Summary

`#[data(impl)] async fn` is not currently supported for records or data enums.

This is not a C#-only issue, and it is not just a backend parity issue. On current `main`, this fails before bindings are generated because the Rust macro path treats async value-type methods like sync methods and expects the declared return value directly instead of a future.

## Example

```rust
#[data]
#[derive(Clone, Copy)]
pub struct Point {
    pub x: f64,
    pub y: f64,
}

#[data(impl)]
impl Point {
    pub async fn async_magnitude(&self) -> f64 {
        (self.x * self.x + self.y * self.y).sqrt()
    }
}
```

This currently fails during Rust compilation with a type mismatch like:

```text
expected `f64`, found future
```

The same shape also fails for data enums:

```rust
#[data]
pub enum Shape {
    Circle { radius: f64 },
    Point,
}

#[data(impl)]
impl Shape {
    pub async fn async_describe(&self) -> String {
        match self {
            Shape::Circle { radius } => format!("circle:{radius}"),
            Shape::Point => "point".to_string(),
        }
    }
}
```

## Current State

There are two separate gaps:

1. The `#[data(impl)]` macro path does not lower async record/data-enum methods correctly.
2. Bindgen currently excludes async value-type methods from the ABI contract.

So the first fix needs to happen in the core macro/ABI layer. After that, each backend can expose the method using its natural async shape.

## Expected Support

Once implemented, async `#[data(impl)]` methods should be supported consistently across generated bindings.

Examples:

- C#: expose as `Task<T>`
- Swift: expose as `async` / `await`
- Kotlin: expose as `suspend`
- Java: expose as `CompletableFuture<T>` or the project's chosen async Java shape
- WASM/TypeScript: expose as `Promise<T>`

## Demo Coverage

The demo currently covers async top-level functions and async class methods, including cases that pass or return records.

It does not cover async methods defined directly on `#[data]` records or data enums.

Add demo coverage for:

1. A `#[data]` record async method returning a direct scalar, for example `Point::async_magnitude(&self) -> f64`.
2. A `#[data]` enum async method returning an encoded value, for example `Shape::async_describe(&self) -> String`.

## Acceptance Criteria

- `#[data(impl)] async fn` compiles for records.
- `#[data(impl)] async fn` compiles for data enums.
- Async value-type methods are represented in the ABI contract instead of being filtered out.
[...]

PR #276, opened 2026-05-05
Add C# async Task support

mhedgpeth:
This is the C# async part of #146.

It adds generated `Task` / `Task<T>` support for the C# backend for the supported async surface today: top-level functions and class methods. The generated C# API has the normal exact-argument overload plus a `CancellationToken` overload.

Another thing that happened here is that my templates started getting very complicated and so I decided to create some helper templates that I could share, so it didn't read as a wall of text. As I looked at the others, it didn't seem like there was a lot of a standard, but my "Plans have AST not Strings" rule leads me in this direction.

## Out of scope

I created #275 that will extend coverage to async `#[data(impl)]` record/enum methods because Java, Swift, and Kotlin do not currently support that value-type async method shape consistently. The goal of this PR is for C# support parity.

Also as a part of #275 we'll want to update the core ABI processing and I felt that was too much for this PR.


engali94 (approved):
Nice!

PR #281, opened 2026-05-06
Add Kotlin Multiplatform support

### LINKED ISSUE #66: Kotlin Multiplatform support
Right now we generate JNI bindings for Kotlin JVM/Android. Adding KMP support would let the same Rust library work on iOS/macOS/Linux via Kotlin/Native. This would make BoltFFI a nice option for teams already using KMP who want to share Rust code across all platforms without maintaining separate Swift/Kotlin bindings.

The good news is cinterop (Kotlin/Native's FFI) just needs a .def file pointing to our C header, which we already generate. So that would be minimal work generate .def files.

Rough plan:
- `kmp` backend
- Generate a `.def` file alongside the header
- Users wire it up in their KMP gradle build

ahmedre:
Fixes #66.


boltffi_cli/src/commands/generate/languages/kotlin_multiplatform.rs
```diff
@@ -0,0 +1,893 @@
...
+};
+use boltffi_bindgen::{KotlinOptions, ir};
+
+use crate::cli::{CliError, Result};
+#[cfg(test)]
+use crate::commands::generate::generator::SourceCrate;
+use crate::commands::generate::generator::{GenerateRequest, LanguageGenerator, ScanPointerWidth};
+use crate::config::{FactoryStyle as ConfigFactoryStyle, Target};
+
+pub struct KotlinMultiplatformGenerator;
```
engali94: nit: we can rename `kotlin_multiplatform.rs` to `kmp.rs` and `KotlinMultiplatformGenerator` to `KMPGenerator` reads shorter and nicer

boltffi_bindgen/src/render/kmp/mod.rs
```diff
@@ -0,0 +1,893 @@
...
+                .map(boltffi_bindgen::load_library_name),
+            desktop_loader: KotlinDesktopLoader::Bundled,
+        }
+    }
+
+    fn package_path(package_name: &str) -> PathBuf {
+        package_name.split('.').collect()
+    }
+
+    fn render_common_and_actual(
```
engali94, line 121: I don't think this belongs to cli crate
we can open a new backend in boltffi_bindgen and use it similar to other backends.
CLI will consume the the output of the generator and handle packing

engali94 (approved):
Looks good 🎉

PR #282, opened 2026-05-06
Support C# callbacks

mhedgpeth:
This PR adds C# callback support as a part of #146.

It covers callback traits, closures, and async callback traits in the generator and demo tests. I also did quite a bit of work keeping with the design to have lower only know about AST (not Strings) and ensuring String-based stuff is only in the AST or the Template layers

### DIFF
--- a/boltffi_bindgen/src/render/csharp/lower/callbacks.rs
+++ b/boltffi_bindgen/src/render/csharp/lower/callbacks.rs
@@ -223,20 +223,12 @@ impl<'a> CSharpLowerer<'a> {
         for param in &params {
             native_params.extend(param.native_params());
         }
-        let marshals_return_bool = matches!(
-            ret,
-            CallbackReturn::Direct {
-                marshals_bool: true,
-                ..
-            }
-        );
 
         CSharpClosureMethodPlan {
             return_type: self.callback_public_return_type(&method.returns),
             public_params: self.public_param_plans(&method.params),
             native_return_type,
             native_params,
-            marshals_return_bool,
             bridge_params: params.clone(),
             invoke: self.closure_invoke(&ret, &params),
         }
--- a/boltffi_bindgen/src/render/csharp/lower/callbacks.rs
+++ b/boltffi_bindgen/src/render/csharp/lower/callbacks.rs
@@ -286,7 +286,7 @@ impl<'a> CSharpLowerer<'a> {
             CallbackReturnMode::CallbackVtable,
         );
         if method.is_async() {
-            CSharpCallbackEntryPlan::Async(CSharpAsyncCallbackEntryPlan {
+            CSharpCallbackEntryPlan::Async(Box::new(CSharpAsyncCallbackEntryPlan {
                 native_params: self.async_callback_native_params(&method_name, &params),
                 bridge_params: params.clone(),
                 decoded_args: decoded_arg_list(&params),
@@ -295,15 +295,15 @@ impl<'a> CSharpLowerer<'a> {
                 faulted_completion: self.async_fault_completion(&ret),
                 success_completion: self.async_success_completion(&ret),
                 catch_completion: self.async_failure_completion(&ret),
-            })
+            }))
         } else {
             let success = self.sync_callback_success(&method_name, &params, &ret);
-            CSharpCallbackEntryPlan::Sync(CSharpSyncCallbackEntryPlan {
+            CSharpCallbackEntryPlan::Sync(Box::new(CSharpSyncCallbackEntryPlan {
                 native_params: self.sync_callback_native_params(&params, &ret),
                 out_initializer: self.sync_out_initializer(&ret),
                 bridge_params: params,
                 success,
-            })
+            }))
         }
     }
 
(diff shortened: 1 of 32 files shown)


boltffi_bindgen/src/render/csharp/templates/render_csharp/native.txt
```diff
@@ -14,6 +14,24 @@
...
         public UIntPtr align;
+
+        internal static unsafe FfiBuf FromBytes(byte[] bytes)
+        {
+            if (bytes.Length == 0)
+            {
+                return default;
+            }
+
+            void* allocated = NativeMemory.Alloc((nuint)bytes.Length);
```
engali94, line 25: won't that cause lifetime problems? because FfiBuf is Rust owned and it is freed with rust's allocator

boltffi_bindgen/src/render/csharp/templates/render_csharp/callback_bridge.txt
```diff
@@ -0,0 +1,517 @@
...
+{%- when CSharpAsyncCallbackFailurePlan::Direct with { default_value } %}
+                callback(callbackData, {{ default_value }}, new FfiStatus { code = 100 });
+{%- when CSharpAsyncCallbackFailurePlan::Encoded %}
+                callback(callbackData, IntPtr.Zero, UIntPtr.Zero, new FfiStatus { code = 100 });
+{%- endmatch %}
+            }
+        }
+{%- endmatch %}
+{% endfor %}
+        private sealed class Proxy : {{ callback.public_name }}
```
engali94: isn't this going to leak until GCed? because seems that proxy holds the native handle privately and only frees it in the finalizer but the public api returns the interface not the proxy itself so callers cant dispose explicitly, should returned callbacks have an owning wrapper?

boltffi_bindgen/src/render/csharp/lower/callbacks.rs
```diff
@@ -0,0 +1,1574 @@
...
+            return_type: self.callback_public_return_type(&method.returns),
+            is_async: method.is_async(),
+            public_params: self.public_param_plans(&method.params),
+            entry: self.lower_callback_entry(method, abi_method),
+            proxy: self.lower_callback_proxy(method, abi_method),
+            delegates: self.lower_callback_delegates(method, abi_method),
+        }
+    }
+
+    fn lower_closure_method(
```
engali94, line 235: this feels a bit off. 4 paths rebuilding the same shape from method + abi_method.,could we lower once and project into the entry/proxy/delegate/public plans? otherwise adding an ABI case means hitting all four and easy to miss one

boltffi_bindgen/src/render/csharp/templates/render_csharp/callback_bridge.txt
```diff
@@ -0,0 +1,517 @@
...
+                            {
+                                BoltFFICallbackReturn.Free(__boltffiErrorOutPtr);
+                            }
+                            return;
+                        }
+{%- if let Some(fallback) = fallback %}
+{%- match fallback %}
+{%- when CSharpAsyncCallbackFailurePlan::Void %}
+                        callback(callbackData, new FfiStatus { code = 100 });
+{%- when CSharpAsyncCallbackFailurePlan::Direct with { default_value } %}
```
engali94: could we pull it into a shared block seems repeated 5 times?

engali94 (approved):

boltffi_bindgen/src/render/csharp/templates/render_csharp/native.txt
mhedgpeth: Good catch. From my research, the allocator mismatch here points to a broader native inline-closure ABI issue rather than something C# should patch locally.

The clean fix would be to change `boltffi_ffi_rules/src/lib.rs` so `ReturnInvocationContext::InlineClosure` + `ReturnPlatform::Native` + buffer-backed returns uses `WriteToOutBufferParts` instead of `DirectReturn`. Then the foreign closure bridge writes `(ptr, len)` and the generated Rust closure adapter decodes and frees it, matching the existing callback-return convention.

That would affect Swift too, since Swift currently has the same direct-return `FfiBuf` shape for encoded inline closure returns.

Do we want to make that broader ABI correction in this PR, or keep this PR scoped and split the shared native inline-closure return fix into a follow-up? I prefer the former but hesitate to make core changes without your involvement.

boltffi_bindgen/src/render/csharp/templates/render_csharp/callback_bridge.txt
mhedgpeth: Good catch, I exposed the proxy in a non-nested way, added disposable (on demand or GCed) functionality to it, and the Program.rs now uses that with the `using` statement.

boltffi_bindgen/src/render/csharp/lower/callbacks.rs
mhedgpeth: Good call I concolidated this into one

boltffi_bindgen/src/render/csharp/templates/render_csharp/callback_bridge.txt
mhedgpeth: I consolidated this

boltffi_bindgen/src/render/csharp/templates/render_csharp/native.txt
engali94: I agree let's keep it for a separate pr

mhedgpeth:
@engali94 thanks as usual for your detailed reviews. I was able to easily address most of it, but didn't want to overreach on the first one. Take a look at that comment and let me know what you think.

boltffi_bindgen/src/render/csharp/templates/render_csharp/native.txt
mhedgpeth: Great this is #283

PR #284, opened 2026-05-08
Support C# streams

mhedgpeth:
This PR adds C# streams as a part of #146, following the stream work from #114.

It exposes stream subscriptions as idiomatic `IAsyncEnumerable<T>` for C#, covering the async, batch, and callback stream modes through the same public shape. I also kept the renderer split where I want it: lower carries semantic stream facts, and the C#-specific iterator shape stays in the template layer.

One thing I tightened up after review: non-blittable stream items now fail loudly during C# lowering instead of pretending there is a wire-buffer `pop_batch` ABI that the macro does not actually emit today.

The C# demo now covers the Java stream scenarios, plus `SubscribePoints()` for record stream items, plus cancellation and early `await foreach` break cleanup.


mhedgpeth:
@engali94 after this, just the docs are left and then we're done! Thanks for all your help

engali94 (approved):
Well done!

PR #285, opened 2026-05-09
Document C# support

mhedgpeth:
This adds C# to the documentation now that it's ready to be released. I also listed out the languages as I saw the single sentence list of language support.

### DIFF
--- a/docs/src/content/docs/getting-started.mdx
+++ b/docs/src/content/docs/getting-started.mdx
@@ -79,20 +79,8 @@ This creates:
 - `dist/apple/` - XCFramework + Swift bindings
 - `dist/android/` - JNI libraries + Kotlin bindings
 - `dist/java/` - JNI library + Java bindings
+- `dist/csharp/` - NuGet package + native runtime assets
 - `dist/wasm/` - WASM module + TypeScript bindings
-
-C# currently uses a generate-and-integrate flow rather than `pack`:
-
-```bash
-cargo build
-boltffi generate csharp
-```
-
-This creates `dist/csharp/` with generated C# source bindings.
-
-`pack` is for targets where BoltFFI creates a platform package, such as an XCFramework,
-jniLibs/JNI output, or an npm package. C# already has generated bindings, but BoltFFI leaves the
-.NET project or NuGet layout to your app, so the command is `generate csharp`.
 
 To build for a single platform:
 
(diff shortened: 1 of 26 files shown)


mhedgpeth:
@engali94  I had some extra time this evening so I decided to chain these; this is the documentation needed for C#, chained to the streaming feature. After we approve and merge these, all that's left is turning it on in the CLI I believe.

docs/src/content/docs/getting-started.mdx
```diff
@@ -127,6 +142,25 @@ System.out.println(distance(p1, p2));  // 5.0
 
 On Java 16+ with `min_version = 16` in `boltffi.toml`, records are used instead of classes. On Java 8+, the same API works with generated final classes.
 
+## Use in C#
+
+Build the host dynamic library, generate the C# bindings, and include the generated `.cs` files in your .NET project:
+
+```bash
+cargo build
+boltffi generate csharp
```
engali94: should we say `boltffi pack csharp` to be truly ready to use?

docs/src/content/docs/packaging.mdx
```diff
@@ -470,6 +486,69 @@ tasks.test {
 }
 ```
 
+<h2 id="c-sharp-generation">C# Generation</h2>
+
+C# uses a source-generation flow because BoltFFI does not create a .NET project or NuGet package.
```
engali94: is this the standard way to ship .net packages? for context, on apple (and similarly others) we ship a Package.swift + XCFramework (static lib, C header, module map) and users pull it in like any other dep. boltffi pack apple produces all of that in one shot, should pack do the same on the c# side?

engali94 (approved):
Nice!

could you alos update the overview SVG to include the C# block

docs/src/content/docs/getting-started.mdx
mhedgpeth: Yeah I'll do that now on another PR, and then update this one

docs/src/content/docs/packaging.mdx
mhedgpeth: Changing this right now with the updated docs

mhedgpeth:
@engali94 this is rebased, I updated the SVG, everything is back to the `pack` command since I added it.

After we merge this I'll do the last bit which is to flip the CLI and then we can see when to cut a release.

engali94 (approved):
NICE!

PR #286, opened 2026-05-09
Support C# Pack

mhedgpeth:
This PR adds C# package support as a part of #146.

It adds `boltffi pack csharp`, which generates the C# bindings into a packable project layout, builds the Rust native library for the configured .NET runtime identifiers, places those native assets under `runtimes/<rid>/native/`, and emits a local NuGet package that consumers can restore from a local package source.

I also wired C# into `pack all` when `targets.csharp.enabled = true`, added the C# package config fields, and updated the C# demo to consume the generated NuGet package instead of compiling the generated `.cs` files directly.

One thing I kept explicit here is that this is normal C# support, not experimental support. This is the last C# release step, so I did not add an experimental flag or language around it.

I also kept the generated `.csproj` as a small Askama template to avoid explicit XML escape code. The pack layer is mixed today, but for this XML-shaped package manifest it made the final file easier to review and kept escaping centralized.

### DIFF
--- a/BOLTFFI_TOML_SPEC.md
+++ b/BOLTFFI_TOML_SPEC.md
@@ -357,6 +357,30 @@ Controls npm package generation in `boltffi pack wasm`.
 - `repository` (string, optional): Package repository URL.
   - Default: `{package.repository}`
 
+## C#
+
+### `[targets.csharp]` (optional)
+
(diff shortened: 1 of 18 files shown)


mhedgpeth:
@engali94 I'll merge this, then update #285 to align, then will flip the experimental flag off.

engali94 (approved):

PR #287, opened 2026-05-09
Add boltffi binding crate

engali94:


boltffi_binding/src/ir/ids.rs
```diff
@@ -0,0 +1,115 @@
...
+/// id with the same numeric value never collide.
+///
+/// # Example
+///
+/// `DeclId::Record(RecordId::from_raw(0))` and
+/// `DeclId::Function(FunctionId::from_raw(0))` share the same raw value but
+/// are distinct `DeclId`s; both can coexist in the same contract.
+#[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd, Serialize, Deserialize)]
+#[non_exhaustive]
+pub enum DeclId {
```
mhedgpeth: I know that this fits the pattern of everything else but I had to look at it twice because the `l` doesn't go with the `Id` - I wouldn't mind just using `Declaration` everywhere but I tend to like longer names.

boltffi_binding/src/ir/metadata.rs
```diff
@@ -0,0 +1,204 @@
...
+}
+
+/// Source-derived metadata attached to a top-level declaration.
+///
+/// Documentation and deprecation are the two facts the binding contract
+/// preserves at declaration scope. Everything else (source spans,
+/// attributes, attribute arguments) is consumed by the classifier or
+/// dropped.
+#[derive(Clone, Debug, Default, Eq, Hash, PartialEq, Serialize, Deserialize)]
+pub struct DeclMeta {
```
mhedgpeth, line 150: Again everything is straightforward then I look at `DeclMeta` - perhaps an example here would be good?

mhedgpeth (approved):
This looks really good, I had a couple of comments that are hopefully helpful. I am excited to see the first language you integrate this into so I can see it fit together.

boltffi_binding/src/ir/ids.rs
engali94: Nice catch that makes total sense!

mhedgpeth (approved):

PR #297, opened 2026-05-11
Fix shared ref of class type incorrectly wire-encoded across FFI

### LINKED ISSUE #299: Class type shared references incorrectly wire-encoded across FFI boundary
## Description

Shared references of class types (e.g. `&Counter`) are incorrectly routed through wire-encoding when crossing FFI boundaries. 

## Reproduction

```rust
pub struct User {
    name: String
}

#[export]
pub fn greet(user: &User) {
    println!("Hello {}", user.name);
}
```

This produces a compile-time error in the generated FFI shim, the macro attempts to decode `&Counter` from a byte buffer instead of passing the pointer through. This works fine in `v0.22.1` and before, it's only broken starting in `v0.23.0`.

phongddo:
## Description

Fixes https://github.com/boltffi/boltffi/issues/299

Shared references of class types (e.g. `&Counter`) crossing FFI boundaries broke starting in `v0.23.0`.

## How to reproduce

```rust
pub struct User {
    name: String
}

#[export]
pub fn greet(user: &User) {
    println!("Hello {}", user.name);
}
```

This works fine in `v0.22.1` and before.

## Root cause

https://github.com/boltffi/boltffi/pull/131  introduced a catch-all [here](https://github.com/boltffi/boltffi/pull/131/changes#diff-b1e2892560c7b040550b0518b1a61cd653026b65447995fb02be41a0f5b54a4dR318-R329) to support borrowed wire-encoded data type params (e.g. `&UserProfile`). However, this was too broad, it also matched `&Counter` where Counter is a class type, routing it through wire-encoding instead of pass-through.

## What's changed

- Class types are neither data types nor custom types, so `named_type_category(&Counter)` should return `None`. The fix narrows the guard to only wire-encode references to known named data types. 
- Updated tests so it registers data types (e.g. `UserProfile`, `Filter`) in `ReturnLoweringContext`.

### DIFF
--- a/boltffi_macros/src/data/expansion/record_impl.rs
+++ b/boltffi_macros/src/data/expansion/record_impl.rs
@@ -643,7 +643,7 @@ mod tests {
     use super::*;
     use crate::index::callback_traits::CallbackTraitRegistry;
     use crate::index::custom_types::CustomTypeRegistry;
-    use crate::index::data_types::DataTypeRegistry;
+    use crate::index::data_types::{DataTypeCategory, DataTypeRegistry};
     use crate::lowering::returns::model::ReturnLoweringContext;
 
     fn parse_impl(code: &str) -> syn::ItemImpl {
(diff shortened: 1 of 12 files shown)


engali94:
Nice catch @phongddo thanks for fixing it
one small thing could you extend the demos to practise this path so the CI would pass.

engali94 (approved):

Issue #320, opened 2026-05-18
Java/C# bindgen drops exports whose signatures use built-in value types (Duration, SystemTime, URL, UUID)


mhedgpeth:
I hit this while filling Java demo coverage gaps for #309, then again while
filling C# coverage gaps. The `builtins` module in `examples/demo` defines ten
exports (`echo_duration`, `make_duration`, `duration_as_millis`,
`echo_system_time`, `system_time_to_millis`, `millis_to_system_time`,
`echo_url`, `url_to_string`, `echo_uuid`, `uuid_to_string`). Kotlin — same JVM —
generates idiomatic wrappers for all ten on `java.time.Duration`,
`java.time.Instant`, `java.net.URI`, and `java.util.UUID`. Java and C# both
emit none of them, with no warning or error.

```bash
cd examples/demo
cargo run -p boltffi_cli -- generate java --output /tmp/java_gen
grep -E 'echoDuration|echoSystemTime|echoUrl|echoUuid' /tmp/java_gen/com/boltffi/demo/Demo.java
# (no output)

cargo run -p boltffi_cli -- generate csharp --output /tmp/csharp_gen
grep -E 'EchoDuration|EchoSystemTime|EchoUrl|EchoUuid' /tmp/csharp_gen/Demo.cs
# (no output)
```

The ten cases were originally catalogued as Java/C# coverage gaps in #309. I'm
reclassifying them to implementation gaps on the respective coverage-gap
branches since there's nothing to write coverage against yet.

PR #330, opened 2026-05-18
Fix java builtin value support

osa8bit:
Handles JAVA part of #320


mhedgpeth (approved):
This looks good let's keep the issue #320 open I can fix the C# side on another PR.

PR #375, opened 2026-05-28
Fix Apple modulemap discovery

### LINKED ISSUE #374: pack apple hides module.modulemap from Xcode 26 explicit module scanner
## Summary

`boltffi pack apple` produces xcframeworks whose `module.modulemap` lives under a namespaced subdirectory (e.g. `Headers/shared/module.modulemap`). Xcode 26 enables explicit module builds by default, and its scanner only inspects the declared `HeadersPath` (`Headers/`) without recursing, so the modulemap is never discovered and the Swift compiler fails with `unable to resolve module dependency`.

Discovered while fixing the Apple build for crux's counter example: redbadger/crux#539. That PR patches the modulemap back into place from a Justfile as a stopgap; the fix belongs here.

## Steps to reproduce

1. Run `boltffi pack apple` against any project (e.g. crux's `examples/counter`).
2. Inspect the generated xcframework:
   ```
   Shared.xcframework/ios-arm64/Headers/
   └── shared/
       ├── module.modulemap
       └── shared.h
   ```
3. Consume the xcframework from a Swift target built with Xcode 26 (explicit modules on by default), e.g. `import SharedFFI`.

## Expected

The Swift compiler resolves `SharedFFI` and the build succeeds.

## Actual

```
error: unable to resolve module dependency: 'SharedFFI'
import SharedFFI
       ^
```

The scanner searches `Headers/` (per `Info.plist`'s `HeadersPath`) for `module.modulemap`, does not recurse into `shared/`, and reports the module as missing.

## Root cause

In `boltffi_cli/src/pack/apple/xcframework.rs`, `prepare_headers` writes `module.modulemap` at the root of the headers staging directory (line 230). `HeaderNamespace::apply_to_xcframework` (lines 252 to 298) then moves every entry under `Headers/`, including `module.modulemap`, into the namespace subdirectory. The `namespaces_slice_headers` test (line 466) currently asserts this nested placement.

## Suggested fix

After namespacing, leave a top-level `module.modulemap` at `Headers/` that references the namespaced header:

```
module SharedFFI {
    header "shared/shared.h"
    export *
}
```

This is what crux#539 patches in by hand and is the minimal change. The alternative, keeping the modulemap at the root from the start and only namespacing `.h` files, is cleaner but breaks the `keeps_namespaced_headers_stable` invariant.

mhedgpeth:
This closes #374.

It keeps the headers namespaced under `Headers/<library>/`, but leaves the modulemap at `Headers/module.modulemap` so Xcode 26's explicit module scanner can discover the module.

This should not be a breaking change for supported consumers because the xcframework's declared `HeadersPath` is still `Headers`, the public module remains `<Name>FFI`, and the modulemap points at the same namespaced header path. I also avoid keeping a second nested modulemap because SwiftPM treats that as a duplicate module definition.

Validated with:

- `cargo test -p boltffi_cli pack::apple::xcframework`
- `swift test --package-path examples/platforms/apple`

### DIFF
--- a/boltffi_cli/src/pack/apple/xcframework.rs
+++ b/boltffi_cli/src/pack/apple/xcframework.rs
@@ -200,7 +200,7 @@ impl<'a> XcframeworkBuilder<'a> {
             });
         }
 
-        HeaderNamespace::for_library(self.config.library_name())
+        HeaderNamespace::new(self.config.library_name(), self.config.xcframework_name())
             .apply_to_xcframework(&xcframework_path)?;
 
         Ok(xcframework_path)
@@ -225,8 +225,10 @@ impl<'a> XcframeworkBuilder<'a> {
 
         copy_directory_contents(&self.headers_dir, &headers_staging)?;
 
-        let modulemap_content =
-            generate_modulemap(&self.config.xcframework_name(), self.config.library_name());
+        let modulemap_content = generate_modulemap(
+            &self.config.xcframework_name(),
+            &format!("{}.h", self.config.library_name()),
+        );
         let modulemap_path = headers_staging.join("module.modulemap");
 
         fs::write(&modulemap_path, modulemap_content).map_err(|source| CliError::WriteFailed {
@@ -240,12 +242,14 @@ impl<'a> XcframeworkBuilder<'a> {
 
 struct HeaderNamespace {
     directory_name: String,
+    module_name: String,
 }
 
 impl HeaderNamespace {
-    fn for_library(library_name: impl Into<String>) -> Self {
+    fn new(library_name: impl Into<String>, module_name: impl Into<String>) -> Self {
         Self {
             directory_name: library_name.into(),
+            module_name: module_name.into(),
         }
     }
 
@@ -294,7 +298,9 @@ impl HeaderNamespace {
 
         entries
             .into_iter()
-            .try_for_each(|source_path| self.move_entry(source_path, &namespace_path))
+            .try_for_each(|source_path| self.move_entry(source_path, &namespace_path))?;
+
+        self.write_root_modulemap(headers_path)
     }
 
     fn move_entry(&self, source_path: PathBuf, namespace_path: &Path) -> Result<()> {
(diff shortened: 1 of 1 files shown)


engali94 (approved):
Nice catch!

engali94:
I tested it with two local binary targets, each shaped like this:

`Headers/module.modulemap`
`Headers/mylib/mylib.h`

if we use a plain `swift build` it will work but using `xcodebuild` on Xcode 26.0.1 failed before compilation because both XCFrameworks produced the same build output: `Build/Products/Debug/include/module.modulemap`
it errored here because Xcode flattens both header directories (from libA and libB) into the same `include/` directory.

So the module names do not collide, but the raw static library XCFramework header processing does. 

I also tried pointing `HeadersPath` at `Headers/<library>` and keeping `module.modulemap` there but Xcode still flattened it to `include/module.modulemap`.

Also I tried named module maps like `DemoFFI.modulemap` it avoided the filename collision but were not discovered at all, Xcode can'r see them.

if we look at An XCFramework slice content can be either:
```
Demo.xcframework/
    libdemo.a + Headers/...
```

Or

```
Demo.xcframework/
   Demo.framework/
```

so trying the second layout solved both the collision and discoverability on xcode 26

we can change our packaging to be something like this:

```txt
Demo.xcframework
  ios-arm64/
    DemoFFI.framework/
      DemoFFI
      Headers/
      Modules/module.modulemap

  ios-arm64_x86_64-simulator/
    DemoFFI.framework/
      DemoFFI
      Headers/
      Modules/module.modulemap
```

so currently the flow is roughly something like this:

```
read boltffi config
for each configured Apple target:
  cargo build Rust staticlib for that target
combine simulator libs with lipo if needed
call xcodebuild -create-xcframework with libdemo.a + Headers
```

after the change it would be:

```
read boltffi config

for each configured Apple target:
  cargo build Rust staticlib for that target

combine simulator libs with lipo if needed
combine macOS libs with lipo if needed

for each final platform slice:
  create a temporary DemoFFI.framework (whatever the lib name is)
  copy the static lib into DemoFFI.framework/DemoFFI
  copy headers into DemoFFI.framework/Headers/demo/
  write DemoFFI.framework/Modules/module.modulemap
  write DemoFFI.framework/Info.plist

call xcodebuild -create-xcframework with DemoFFI.framework slices
```

mhedgpeth:
@engali94 I fixed it per our conversation, and I have a test in there that tests this, but that's not a part of CI.

PR #385, opened 2026-05-30
Fix Java/C# records with Result fields

### LINKED ISSUE #322: Java/C# bindgen drops records that embed Result<T, E> as a field (BenchmarkResponse)
Found while filling Java demo coverage gaps for #309, then again while
filling C# coverage gaps. `BenchmarkResponse` is

```rust
pub struct BenchmarkResponse {
    pub request_id: i64,
    pub result: Result<DataPoint, ComputeError>,
}
```

with six exports around it (`create_success_response`,
`create_error_response`, `is_response_success`, `get_response_value`, plus
the two `should_report_*_response` cases). Java's and C#'s generated bindings
don't emit `BenchmarkResponse` as a type at all, and every one of those six
exports is silently dropped.

```bash
cd examples/demo
cargo run -p boltffi_cli -- generate java --output /tmp/java_gen
ls /tmp/java_gen/com/boltffi/demo/BenchmarkResponse.java
# ls: ... No such file or directory

cargo run -p boltffi_cli -- generate csharp --output /tmp/csharp_gen
ls /tmp/csharp_gen/BenchmarkResponse.cs
# ls: ... No such file or directory
```

Six `results.error_enums.benchmark_response.*` cases were originally
catalogued as Java/C# coverage gaps in #309. I'm reclassifying them to
implementation gaps on the respective coverage-gap branches.

Probably related to #321 — both look like Java/C# don't handle `Result<T, E>`
outside the return position.

mhedgpeth:
This closes #322.

It stops the Java and C# bindgens from dropping records and data-enum payloads that embed `Result<T, E>` as a field. Both backends now admit nested `Result` shapes, lower them to public `BoltFFIResult` carriers, and decode the tagged wire payload back into the right Ok/Err branch.

The C# runtime helper is now emitted whenever generated public surface area names `BoltFFIResult`, not only when callbacks are present. Java also gives `BoltFFIResult` real equality/hash behavior, including array, optional, and list payloads, so record equality keeps working for nested wire shapes.

I also removed the Java/C# implementation gaps for the `BenchmarkResponse` demo catalogue cases and added matching platform assertions for the six success/error/value cases. Python stays excluded because that support is still separate.

### DIFF
--- a/boltffi_bindgen/src/render/csharp/ast/code.rs
+++ b/boltffi_bindgen/src/render/csharp/ast/code.rs
@@ -141,6 +141,13 @@ pub(crate) enum CSharpExpression {
         type_args: Vec<CSharpType>,
         args: CSharpArgumentList,
     },
+    /// `{receiver}.{method}({arg0}, ...)` where the receiver is a
+    /// generic type expression such as `BoltFFIResult<T, E>`.
+    StaticMethodCall {
+        receiver: CSharpType,
+        method: CSharpMethodName,
+        args: CSharpArgumentList,
+    },
     /// `({target}){inner}`: a C-style cast.
     Cast {
         target: CSharpType,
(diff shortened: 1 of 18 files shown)


mhedgpeth:
@engali94 I want to fit this and the other targeted remaining issue to `0.25.2` before we release that. I'll do them one at a time to avoid confusion/risk of issues.

engali94 (approved):

PR #387, opened 2026-05-30
Fix C# Result parameters

### LINKED ISSUE #321: Java/C# bindgen drops exports that take Result<T, E> as a parameter
Found while filling Java demo coverage gaps for #309, then again while
filling C# coverage gaps. `results::basic::result_to_string` takes
`Result<i32, String>` as a parameter and returns `String`. Java's generated
`Demo.java` and C#'s generated `Demo.cs` don't include `resultToString` /
`ResultToString` at all. Exports that *return* `Result<T, E>` generate fine
(they map to thrown exceptions); only Result-as-parameter is dropped,
silently.

```bash
cd examples/demo
cargo run -p boltffi_cli -- generate java --output /tmp/java_gen
grep resultToString /tmp/java_gen/com/boltffi/demo/Demo.java
# (no output)

cargo run -p boltffi_cli -- generate csharp --output /tmp/csharp_gen
grep ResultToString /tmp/csharp_gen/Demo.cs
# (no output)
```

Two cases (`results.basic.result_to_string.should_render_ok` /
`...should_render_err`) were originally catalogued as Java/C# coverage gaps
in #309. I'm reclassifying them to implementation gaps on the respective
coverage-gap branches.

Probably related to #322 — both look like Java/C# don't handle `Result<T, E>`
outside the return position.

mhedgpeth:
This closes #321.

It stops the C# bindgen from dropping functions that take `Result<T, E>` as a value parameter. The lowerer now admits Result params when both branches are supported, classifies their ABI shape as wire-encoded, and passes the generated byte buffer into the native call instead of trying to marshal the public `BoltFFIResult` carrier directly.

The C# runtime predicate now emits `BoltFFIResult<TOk, TErr>` whenever generated public surface area names it, including Result-param-only modules. I also added renderer coverage for the `Result<i32, String> -> String` shape that triggered the demo gap.

I removed the C# implementation-gap exclusions for `results.basic.result_to_string` and added C# platform assertions for the Ok and Err cases. Python stays excluded because its Result-parameter support is separate.

### DIFF
--- a/boltffi_bindgen/src/render/csharp/emit.rs
+++ b/boltffi_bindgen/src/render/csharp/emit.rs
@@ -1455,6 +1455,50 @@ mod tests {
     }
 
     #[test]
+    fn emit_function_with_result_param_uses_wire_writer_and_result_runtime() {
+        let mut contract = empty_contract();
+        contract.functions.push(function_with_types(
+            "result_to_string",
+            vec![(
+                "input",
+                TypeExpr::Result {
+                    ok: Box::new(TypeExpr::Primitive(PrimitiveType::I32)),
+                    err: Box::new(TypeExpr::String),
+                },
+            )],
+            ReturnDef::Value(TypeExpr::String),
+        ));
+
+        let src = emit_contract(&contract).combined_source();
+
+        assert_source_contains(
+            &src,
+            "public static string ResultToString(BoltFFIResult<int, string> input)",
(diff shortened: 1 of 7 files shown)


mhedgpeth:
@engali94 this should complete all the issues for the `0.25.2` milestone, I hope we can release that after this is done.

engali94 (approved):

PR #394, opened 2026-05-31
track unsupported scanner cases

engali94:


mhedgpeth (approved):
this looks good it might be interesting to create a negative test demo or set of rust files/negative demo crates that would prove out what is not supported at the integration level (and act as some form of documentation too)

engali94:
> this looks good it might be interesting to create a negative test demo or set of rust files/negative demo crates that would prove out what is not supported at the integration level (and act as some form of documentation too)

Nice idea! I added them

PR #399, opened 2026-05-31
fix: update docs to include reference to current boltffi version

jottinger:


mhedgpeth (approved):
This fixes the issue, approved - @engali94 not sure if you want this to be templated with "latest version" somehow? That might be overthinking it...

engali94:
> This fixes the issue, approved - @engali94 not sure if you want this to be templated with "latest version" somehow? That might be overthinking it...

I am thinking of making the CI pump the version every time we have a new release

engali94:
@mhedgpeth I think this is good enough to merge. Then we can agree on how to keep the docs site and README versions in sync with the latest release. What do you think?

PR #516, opened 2026-06-10
IR based expansion for records

engali94:


boltffi_macros/src/experimental/target/closure_registration.rs
```diff
@@ -0,0 +1,94 @@
+use boltffi_binding::{ImportSymbol, Native, NativeSymbol, Surface, Wasm32, native, wasm32};
+
+use crate::experimental::error::Error;
+
+/// A render lane for a foreign-provided closure registering with Rust.
+pub enum IncomingClosureLane<'a> {
```
mhedgpeth: My personal style is I prefer lifetimes to be named. Where do we expect `'a` to come from? Could we name it that?

boltffi_macros/src/experimental/wrapper/encoded/custom.rs
```diff
@@ -249,12 +249,71 @@ impl<'context, 'a, S: Target> Outgoing<'context, 'a, S> {
...
+                    true => Error::UnsupportedExpansion("custom conversion inside map codec"),
+                    false => Error::UnsupportedExpansion("map codec conversion"),
+                },
+            ),
+            _ => Ok(value),
+        }
+    }
+}
+
+impl<'context, 'a, S: Target> BorrowedOutgoing<'context, 'a, S> {
```
mhedgpeth: here one thing is named and the other isn't, which kind of destroys the meaning - 'context is for what? 'a is for what? I can't tell by reading this

'a is the CodecNode 'copntext is the Expansion which has an 'a - what are these actual referred sources?

mhedgpeth (approved):

mhedgpeth (commented):
Much more readable with the named lifetimes; thanks!

PR #517, opened 2026-06-10
support class handle returns

### LINKED ISSUE #257: Support instance methods returning class handles
Constructors can return class handles today, but ordinary methods returning class objects do not work.
```rust
#[export]
impl Counter {
    pub fn snapshot(&self) -> Counter {
        Counter::new(self.get())
    }
}
```

This currently falls through normal return lowering and gets treated as wire-encoded data, so the macro tries to require `Counter: WireEncode`. That is the wrong transport for classes. A returned class should cross the boundary as an object handle, same as constructors.

Expected behavior:

```rust
pub fn snapshot(&self) -> Counter
pub fn maybe_snapshot(&self) -> Option<Counter>
```

should return object handles, with `None` represented as a null handle.

This is a related to #256. #256 covers `Option<Self>` constructors and methods that consume `self`. This issue is about non-consuming methods returning object handles.

engali94:
fixes #257

### DIFF
--- a/boltffi_bindgen/src/render/c/lower.rs
+++ b/boltffi_bindgen/src/render/c/lower.rs
@@ -49,7 +49,6 @@ impl<'a> CHeaderLowerer<'a> {
         .render()
         .unwrap();
 
-        out.push_str(&self.forward_declarations());
         out.push_str(&self.composite_struct_typedefs());
         out.push_str(&self.enum_typedefs());
         out.push_str(&self.callback_vtables());
@@ -59,26 +58,6 @@ impl<'a> CHeaderLowerer<'a> {
         out.push_str(&self.free_functions());
 
         out
-    }
-
-    fn forward_declarations(&self) -> String {
-        let names: Vec<_> = self
-            .contract
-            .catalog
-            .all_classes()
-            .map(|c| c.id.as_str().to_string())
-            .collect();
-
-        if names.is_empty() {
-            return String::new();
-        }
-
-        let decls: String = names
-            .iter()
-            .map(|name| format!("struct {};\n", name))
-            .collect();
-
-        format!("\n{}\n", decls)
     }
 
     fn composite_struct_typedefs(&self) -> String {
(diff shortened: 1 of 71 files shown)


osa8bit (approved):

boltffi_bindgen/src/render/c/lower.rs
```diff
@@ -62,23 +62,7 @@ impl<'a> CHeaderLowerer<'a> {
...
-            return String::new();
-        }
-
-        let decls: String = names
-            .iter()
-            .map(|name| format!("struct {};\n", name))
-            .collect();
-
-        format!("\n{}\n", decls)
+        String::new()
```
mhedgpeth: kind of weird to read this, it reads like it's a mistake

examples/demo/src/classes/unsafe_single_threaded.rs
```diff
@@ -1,7 +1,73 @@
...
+        }
+    }
+
+    #[demo_bench_macros::demo_case(
+        "classes.unsafe_single_threaded.map_view.add_marker.should_return_single_threaded_marker_handle",
+        justification = "Ensure a single-threaded class method returns another single-threaded class as an owned object handle.",
+        directions = "Call `classes::unsafe_single_threaded::MapView::add_marker` through the generated binding and assert the returned Marker handle exposes the MarkerOptions fields.",
+        exclude(
+            csharp,
+            reason = ExclusionReason::ImplementationGap,
```
mhedgpeth: should we/could we create follow up tickets for these and refer to these by the ticket #?

mhedgpeth (approved):
I'm still learning the overall plot here so a lot of this is me checking over the code, and what I know.

I also will say that it makes me somewhat uncomfortable to see the same quoted string spread out a lot, like "*void" - but it's the crate convention so I left it out of the inline comments.

boltffi_bindgen/src/render/c/lower.rs
engali94: nice catch, fixed now

examples/demo/src/classes/unsafe_single_threaded.rs
engali94: turned out thefix was simple

PR #518, opened 2026-06-10
Port Ruby generation to the IR backend

### LINKED ISSUE #48: Add Ruby language bindings generation
Ruby calls C functions in two main ways, the first one is native C extensions via Ruby's C API (`rb_define_method`, `VALUE` types) or FFI gem for dynamic loading without compilation both solution need to be MVPed first and pick the most preferment one with best ergonomics

ianks:
## Summary

This PR adds the experimental Ruby target on the IR backend path.

Ruby generation now uses `boltffi_binding` + `boltffi_backend` with the shared C ABI bridge, and emits a source-gem-shaped native extension package (`_native.c`, `boltffi.h`, `extconf.rb`, `lib/<crate>.rb`, and gemspec). The generated extension links against the Rust `staticlib`, so the first Ruby slice can stay close to the existing bridge instead of introducing a separate FFI loading path.

## What’s supported

The current Ruby backend supports sync free functions for:

- primitives
- strings and bytes
- mutable byte buffers
- direct/blittable record returns
- encoded record returns for native field shapes like `String`, `Bytes`, and primitives

Record objects are generated as Ruby `TypedData` classes. Direct records wrap the C record typedef directly, while encoded record returns decode `FfiBuf_u8` into native C field storage and expose Ruby getters.

A few details are intentionally conservative:

- record classes are return-only for now, so allocation is explicitly undefined with `rb_undef_alloc_func`
- encoded records store native C/Rust field representations, not Ruby `VALUE`s
- decode paths protect Ruby allocation and free `FfiBuf_u8` buffers before re-raising
- Ruby-head ABI markers are preserved/exported when `RUBY_DLN_CHECK_ABI` is present
- `RUBY_TYPED_EMBEDDABLE` follows Ruby’s TypedData pattern, with direct records using `RUBY_TYPED_DEFAULT_FREE` and encoded records freeing only auxiliary buffers on embeddable Rubies

## Current scope

Ruby is still experimental. This does not add callbacks, streams, async methods, record parameters, data enums, or associated record methods yet.

Unsupported declarations return typed backend diagnostics instead of generating broken C.

## Validation

- `cargo fmt --all -- --check`
- `cargo test -p boltffi_backend ruby`
- `cargo test -p boltffi_cli ruby`
- `ruby -c examples/platforms/ruby/run_tests.rb`
- `./examples/platforms/ruby/test-demo.sh`
- `./examples/demo/verify-platform-demos.sh --platform ruby`
- Built Ruby from `/Users/ianks/src/github.com/ruby/ruby` with `--enable-shared` and verified `./examples/platforms/ruby/test-demo.sh --ruby /tmp/ruby-boltffi-shared/bin/ruby`

closes #48


boltffi_bindgen/src/render/ruby/emit.rs
```diff
@@ -0,0 +1,476 @@
...
+            ),
+            RubyType::Bytes => (
+                "FfiBytes".into(),
+                "bolt_rb_to_ffi_bytes".into(),
+                true,
+                "T_STRING",
+            ),
+            // C-style (repr(i32)) enums cross the ABI boundary as int32_t.
+            RubyType::Enum(_) => ("int32_t".into(), "NUM2INT".into(), false, ""),
+            _ => ("VALUE".into(), "".into(), false, ""),
```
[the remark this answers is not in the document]
ianks: Leaving this open as a tracked gap. In the current implementation, handle params never cross the ABI as raw `VALUE`: `is_supported_param` rejects `TypeExpr::Handle(_)`, so callables with handle params are skipped with a warning rather than generating mismatched glue.

Proper TypedData unwrap (`rb_check_typeddata` → `*mut Class`) is deferred until after this PR lands.

boltffi_bindgen/src/render/ruby/naming.rs
```diff
@@ -0,0 +1,64 @@
...
+            | "if"
+            | "else"
+            | "elsif"
+            | "unless"
+            | "while"
+            | "until"
+            | "for"
+            | "in"
+            | "return"
+            | "yield"
```
tavianator: `yield` (and probably a few others here) is not actually a reserved method name, and for example if you want to implement the `Fiber::Scheduler` interface you'll have to name a method `yield`: https://docs.ruby-lang.org/en/4.0/Fiber/Scheduler.html#method-i-yield

ianks:
Separately from the technical review, I wanted to give a bit of context on my intent here.

I’m interested in helping make Ruby a well-supported BoltFFI target, not just landing this initial PR. I work at Shopify, where Ruby support here has practical value for us, and I maintain [`rb-sys`](https://github.com/oxidize-rb/rb-sys), which is used by a decent chunk of the Rust-backed Ruby native extension ecosystem.

I’m happy to keep owning Ruby-specific follow-up work after this lands: bug fixes, compatibility issues, CI/demo maintenance, and helping shape the target so it fits Ruby conventions without adding undue maintenance burden for BoltFFI maintainers.

No expectation that this changes the bar for the PR, I just wanted to make clear that this isn’t intended as a drive-by contribution.

mhedgpeth:
Hey there I love this! Let me know if you want to talk and I can help you get going. I am in the US and recently did the C# implementation with a lot of lessons learned. You can view that ticket plan to get an idea of a good project plan to follow that will keep you sane. Let me know how I can help! We are on discord too if that works better

ianks:
@mhedgpeth Discord would be great! Can you send an invite to `ianks_shopify`?

engali94:
Thanks for this @ianks. One thing before you take it further. This is built on the old `boltffi_bindgen/src/render/<lang>/` path, and we've frozen that path. New targets go through the IR now.  Instead of pulling transport decisions out of the AST yourself, the IR hands you those decisions already made and you just read declarations and emit source. It's less code and a lot less to keep in your head.

We just wrote the guides for it:
   - [Contributing](https://github.com/boltffi/boltffi/blob/feature/backend-host-bridge/docs/contributors/contributing.md)
   - [Adding a new target language](https://github.com/boltffi/boltffi/blob/feature/backend-host-bridge/docs/contributors/adding-a-target.md)

The second doc walks the whole thing start to finish. Worth following this Ruby's issue https://github.com/boltffi/boltffi/issues/48 

A lot of what you already figured out carries over, so this isn't starting from scratch. Happy to help you scope the port.

Our Discord server: https://discord.com/invite/Q6A7zNNFk3

boltffi_bindgen/src/render/ruby/templates/render_ruby/extconf.rb.txt
```diff
@@ -0,0 +1,197 @@
...
+#     ld64, `--exclude-libs,ALL` on GNU ld).
+#   • Dead-code stripping (`-dead_strip`/`--gc-sections`) then drops the Rust
+#     code unreachable from Init_. Init_ is compiled from this gem's _native.c
+#     (not the archive), so the hiding/stripping never touches it.
+#
+# Branch on the *target* OS (target_os, not host_os): under rake-compiler-dock
+# cross builds, host_os is the build box, not the gem's target.
+case RbConfig::CONFIG["target_os"]
+when /darwin/
+  $LDFLAGS << " -Wl,-exported_symbol,_Init_{{ crate_name }}_native"
```
tenderlove: I'm not sure that we should do this. Edge versions of Ruby want `ruby_abi_version` exported (which happens automatically). This flag prevents it from being exported and the extension won't work.

ianks: done

PR #559, opened 2026-06-18
move the legacy in-progress python to IR-based backend

engali94:


mhedgpeth (approved):
This looks really good, I'm trusting in the tests to ensure it works, and mainly reviewed this for how much it made sense.

The solution is very straightforward, as I read through it I'm rarely confused. Nice work!

The two spots that popped out for me:

1. When to allow `String` in the lower types is definitely a judgement call. The whole problem I ran into when I did this in C# was that lower is essentially a `String` builder in itself; we're generating code. So what you put in these strings is an interesting rule. What I settled on was: I can only put AST elements from the language in these places, and I'll represent these as new types; this way the lower layer is completely straightforward as to the intent, and cheating doesn't happen where I fit "phrases" into the lower layer and have to manage both the template, and sub-phrases within lower...when I fell into that trap things got confusing.

That's a long way to say: for all `String` elements in the lower types, consider a new type that would say what it is, framed in terms that someone would read in a Python book. Or alternatively think "Michael takes things too far" and go with what you have.

2. As I read through it there is (A) a lot of matching logic that felt similar at times, (B) lots of chained iterator stuff. I didn't take the time to know where if anywhere the duplication exists, but there was definitely the smell of the methods being quite long and the cyclomatic complexity being quite high in some of the functions. But I'm torn because they're so easy to read, and we're (as stated above) making Strings here; not curing cancer (directly at least).

So I'm a bit torn but it seemed to me if you went to the high cyclomatic complexity methods in this PR, and resolved the [inappropriate intimacy](https://refactoring.guru/smells/inappropriate-intimacy) smell, you might have a more expressive approach at the end, albeit with the risk of it becoming too abstract. I really do like how straightforward it all is at the moment, but the thought came up.

engali94:
> This looks really good, I'm trusting in the tests to ensure it works, and mainly reviewed this for how much it made sense.
> 
> The solution is very straightforward, as I read through it I'm rarely confused. Nice work!
> 
> The two spots that popped out for me:
> 
> 1. When to allow `String` in the lower types is definitely a judgement call. The whole problem I ran into when I did this in C# was that lower is essentially a `String` builder in itself; we're generating code. So what you put in these strings is an interesting rule. What I settled on was: I can only put AST elements from the language in these places, and I'll represent these as new types; this way the lower layer is completely straightforward as to the intent, and cheating doesn't happen where I fit "phrases" into the lower layer and have to manage both the template, and sub-phrases within lower...when I fell into that trap things got confusing.
> 
> That's a long way to say: for all `String` elements in the lower types, consider a new type that would say what it is, framed in terms that someone would read in a Python book. Or alternatively think "Michael takes things too far" and go with what you have.
> 
> 2. As I read through it there is (A) a lot of matching logic that felt similar at times, (B) lots of chained iterator stuff. I didn't take the time to know where if anywhere the duplication exists, but there was definitely the smell of the methods being quite long and the cyclomatic complexity being quite high in some of the functions. But I'm torn because they're so easy to read, and we're (as stated above) making Strings here; not curing cancer (directly at least).
> 
> So I'm a bit torn but it seemed to me if you went to the high cyclomatic complexity methods in this PR, and resolved the [inappropriate intimacy](https://refactoring.guru/smells/inappropriate-intimacy) smell, you might have a more expressive approach at the end, albeit with the risk of it becoming too abstract. I really do like how straightforward it all is at the moment, but the thought came up.

thanks for the detailed feedback, I handled both points, I borrowed a lot from C# target implementation

PR #561, opened 2026-06-20
add contributor guides for adding a new target

engali94:


mhedgpeth (approved):
This is really good. I wish I would have had this when I did one! A couple of things that might improve it: first, it might be good to tell people to have a master ticket that lays out the approach like we have with C-sharp and Java. Second, it might be good to tell people if a language implementation goes stale that we reserve the right to delete it if it's incomplete and that someone doing a language should complete it in 6 to 8 weeks. I think this would be the least amount of burden on the maintainers and also the sense of urgency actually helps with doing the work because it's pretty complicated to keep the model in your mind at first at least.

engali94:
> This is really good. I wish I would have had this when I did one! A couple of things that might improve it: first, it might be good to tell people to have a master ticket that lays out the approach like we have with C-sharp and Java. Second, it might be good to tell people if a language implementation goes stale that we reserve the right to delete it if it's incomplete and that someone doing a language should complete it in 6 to 8 weeks. I think this would be the least amount of burden on the maintainers and also the sense of urgency actually helps with doing the work because it's pretty complicated to keep the model in your mind at first at least.

good catch! Updated!

PR #574, opened 2026-06-23
Support shallow and deep xcframework based on platform

### LINKED ISSUE #560: `boltffi pack apple` generates shallow bundle for macOS framework slice, breaking macOS builds
When `include_macos = true` is set and `boltffi pack apple` is run, the macOS XCFramework slice (`macos-arm64_x86_64/SharedFFI.framework`) is packaged as a shallow bundle. macOS requires a versioned bundle structure; Xcode rejects the shallow layout with:

```
Framework .../SharedFFI.framework contains Info.plist,
expected Versions/Current/Resources/Info.plist since
the platform does not use shallow bundles
```

**Steps to reproduce**

1. Set `include_macos = true` in `boltffi.toml`
2. Run `boltffi pack apple`
3. Build a macOS target in the generated Xcode project

**Expected behaviour**

The `macos-*` slice should use the versioned bundle layout that macOS requires:

```
macos-arm64_x86_64/SharedFFI.framework/
  Versions/
    A/
      SharedFFI          ← binary
      Headers/
      Modules/
      Resources/
        Info.plist
    Current              → A  (symlink)
  SharedFFI              → Versions/Current/SharedFFI  (symlink)
  Headers                → Versions/Current/Headers    (symlink)
  Modules                → Versions/Current/Modules    (symlink)
  Resources              → Versions/Current/Resources  (symlink)
```

**Actual behaviour**

All slices — including the macOS one — are emitted as shallow bundles:

```
macos-arm64_x86_64/SharedFFI.framework/
  SharedFFI
  Headers/
  Modules/
  Info.plist             ← wrong location for macOS
```

**Context**

iOS, tvOS and watchOS use shallow bundles, so the current output is correct for those slices. macOS is the exception: Apple's bundle programming guide and Xcode's validator both require the versioned layout for macOS frameworks.

This was introduced in v0.25.2 (PR #375) when packaging was switched from raw static libs + headers to framework-based slices. The framework layout is correct for iOS but was applied uniformly to macOS as well.

**Version**

boltffi 0.26.1

**Workaround**

Post-process the macOS slice after `boltffi pack apple` to restructure it into the versioned layout:

```bash
FW="generated/Shared/Shared.xcframework/macos-arm64_x86_64/SharedFFI.framework"
mkdir -p "$FW/Versions/A/Resources"
mv "$FW/SharedFFI"  "$FW/Versions/A/SharedFFI"
mv "$FW/Headers"    "$FW/Versions/A/Headers"
mv "$FW/Modules"    "$FW/Versions/A/Modules"
mv "$FW/Info.plist" "$FW/Versions/A/Resources/Info.plist"
ln -sf A                          "$FW/Versions/Current"
ln -sf Versions/Current/SharedFFI "$FW/SharedFFI"
ln -sf Versions/Current/Headers   "$FW/Headers"
ln -sf Versions/Current/Modules   "$FW/Modules"
ln -sf Versions/Current/Resources "$FW/Resources"
```

**Suggested fix**

In the Apple packaging code, detect when a slice is targeting `macos` and emit the versioned directory structure and symlinks instead of the shallow layout.

mhedgpeth:
Fixes #560 

This is also blocking my own development so I hope we can get a hotfix release out quickly.

### DIFF
--- a/boltffi_cli/src/pack/apple/xcframework.rs
+++ b/boltffi_cli/src/pack/apple/xcframework.rs
@@ -252,6 +252,27 @@ impl AppleLibrarySliceKind {
             Self::MacOs => "macos",
         }
     }
+
+    fn framework_layout(self) -> FrameworkLayout {
+        match self {
+            // iOS, tvOS and watchOS use shallow bundles.
+            Self::IosDevice | Self::IosSimulator => FrameworkLayout::Shallow,
+            // macOS requires the versioned bundle layout; Xcode rejects shallow
+            // macOS frameworks.
+            Self::MacOs => FrameworkLayout::Versioned,
+        }
+    }
+}
+
+/// Layout of a `.framework` bundle.
+///
+/// macOS is the only Apple platform that requires the versioned bundle layout
+/// (`Versions/A` plus `Current` and top-level symlinks); every other platform
+/// uses the shallow layout with resources at the bundle root.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+enum FrameworkLayout {
+    Shallow,
+    Versioned,
 }
 
 #[derive(Debug)]
@@ -333,6 +354,7 @@ impl XcframeworkPlan {
                     library_name.clone(),
                     headers_dir.clone(),
                     library_slice.library_path,
+                    library_slice.kind.framework_layout(),
                 )
             })
             .collect();
@@ -375,6 +397,7 @@ struct StaticFrameworkBundlePlan {
     headers_dir: PathBuf,
     library_path: PathBuf,
     public_header_path: String,
+    layout: FrameworkLayout,
 }
 
 impl StaticFrameworkBundlePlan {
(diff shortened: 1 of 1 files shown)


engali94:
@mhedgpeth sure lets release 0.27.0 right after this getting merged

mhedgpeth:
I tested this locally on my blocking scenario and this works for me, @engali94 I think this is ready to review and then release. Let me know if you want me to do that part; I know there are a lot of moving parts in the project right now with the refactorings/etc.

engali94:
> I tested this locally on my blocking scenario and this works for me, @engali94 I think this is ready to review and then release. Let me know if you want me to do that part; I know there are a lot of moving parts in the project right now with the refactorings/etc.

Nice!
I will check on my end as well, wondering about your use case, why our ci did not catch such failure?

engali94 (approved):

mhedgpeth:
> > I tested this locally on my blocking scenario and this works for me, @engali94 I think this is ready to review and then release. Let me know if you want me to do that part; I know there are a lot of moving parts in the project right now with the refactorings/etc.
> 
> 
> 
> Nice!
> 
> I will check on my end as well, wondering about your use case, why our ci did not catch such failure? 

I didn't have enough time to close that gap but I think it only shows itself when explicitly running the macOS target and compiling that way. So I'm thinking out CI covers the iOS variant

osa8bit (approved):

PR #591, opened 2026-06-26
Migrate Kotlin to use IR based generation

engali94:


boltffi_backend/templates/target/kotlin/enumeration.kt
```diff
@@ -0,0 +1,365 @@
...
+{%- else %}
+{%- for statement in initializer.call() %}
+            {{ statement }}
+{%- endfor %}
+{%- endif %}
+        }
+{%- endfor %}
+{%- for method in enumeration.static_methods() %}
+
+        {% if method.async_call().is_some() %}suspend {% endif %}fun {{ method.name() }}({% for parameter in method.parameters() %}{{ parameter.name() }}: {{ parameter.ty() }}{% if !loop.last %}, {% endif %}{% endfor %}){% if let Some(return_type) = method.returns() %}: {{ return_type }}{% endif %} {
```
wkornewald, line 53: Just a general suggestion to question this approach: I've implemented and worked with a lot of different code generation approaches and transpilers (Python to C++ CUDA transpilers, languages transpiled to JS, multiple OpenAPI and FHIR and XSD and WSDL code generators, ...). Templating based code generation like it's used here (Jinja-inspired) or especially in the openapi-generator project (Mustache-based) quickly become a cause of huge complexity. Non-trivial logic is very hard to follow and debug and lack of auto-complete is a big pain point (not sure if you have solutions for that here?). At least in my experience e.g. rewriting the openapi-generator project (internally) to use normal code-based generation logic with a few nice helpers or macros (e.g. to handle indentation and make writing string-based output with minimal syntax possible so it looks almost as clean as in a template) can simplify the whole logic a lot.

Maybe that's something you want to evaluate in a simple example? Not sure...I just got triggered when I reviewed the code because in my experience template languages just don't scale well when complexity grows beyond very simple HTML output.

Sorry that this doesn't quite belong to this PR, but is a more general discussion. Should this better become a ticket?

osa8bit (approved):

boltffi_backend/templates/target/kotlin/enumeration.kt
engali94: hey @wkornewald thanks for writing this up, good points.

i agree it doesn't fit this PR and I want to hear more about your approaches & experience.

[Let me open a separate discussion so it doesn't get buried here.](https://github.com/boltffi/boltffi/discussions/611#discussion-10368824) Curious how you did the openapi-generator rewrite.

We also have a Discord if you'd rather talk there, either way, let's keep this going

PR #597, opened 2026-07-01
Add guarded KMP IR JVM/Android emission

ahmedre:


boltffi_backend/src/target/kmp/emit/common.rs
```diff
@@ -0,0 +1,64 @@
...
+    ];
+
+    join_kotlin_sections(sections)
+}
+
+fn render_common_result_runtime() -> String {
+    r#"class FfiException(val code: kotlin.Int, message: kotlin.String) : kotlin.Exception(message)
+
+sealed class BoltFFIResult<out T, out E> {
+    data class Ok<T>(val value: T) : BoltFFIResult<T, kotlin.Nothing>()
```
engali94: we should use askam templates for this example https://github.com/boltffi/boltffi/pull/591/changes#diff-13d3835b612142820e8284f16e9b729d7532c7d1c19a90c55af350e9d296213d

ahmedre: done

engali94 (approved):

PR #622, opened 2026-07-07
Fix native OptionScalar exports returning f64 instead of FfiBuf

### LINKED ISSUE #621: `#[export]` fn taking a `#[data]` enum with a data-carrying variant (named-field or tuple) and returning `Option<{integer}>` fails: "expected `FfiBuf`, found `f64`"
### Summary

An `#[export]`-annotated function whose parameter is a `#[data]` enum with a data-carrying variant, and whose return type is `Option<{integer}>`, fails to compile with a nonsensical type error pointing at the `#[export]` attribute itself. This happens whether the variant is named-field (`Pending { count: u8 }`) or tuple (`Pending(u8)`). The function body is irrelevant — it fails even with a trivial `None` body — so this looks like a signature-driven codegen bug rather than anything about the actual logic.

### Environment

- boltffi 0.27.3
- rustc 1.96.0 (ac68faa20 2026-05-25)
- cargo 1.96.0
- Windows

### Minimal reproduction

`Cargo.toml`:
```toml
[package]
name = "boltffi-bug-reproduction"
version = "0.1.0"
edition = "2024"

[dependencies]
boltffi = "0.27.3"
```

`src/lib.rs`:
```rust
#[boltffi::data]
pub enum Signal {
    Pending { count: u8 },
}

#[boltffi::export]
pub fn maybe_count(_signal: Signal) -> Option<u8> {
    None
}
```

`cargo build`:
```
error[E0308]: mismatched types
  --> src\lib.rs:6:1
   |
6  | #[boltffi::export]
   | ^^^^^^^^^^^^^^^^^^
   | expected `FfiBuf`, found `f64`
   | expected `boltffi::__private::FfiBuf` because of return type
   |
   = note: this error originates in the attribute macro `boltffi::export` (in Nightly builds, run with -Z macro-backtrace for more info)
```

The same error occurs with `Pending(u8)` (tuple variant) in place of `Pending { count: u8 }` — the variant's field syntax doesn't matter, only that it carries data.

### Expected

The crate compiles.

### Narrowing down the trigger

I isolated this to the combination of:
- an `#[export]` fn parameter that is a `#[data]` enum containing a variant that carries data — named-field (`Pending { count: u8 }`) or tuple (`Pending(u8)`), both trigger it
- an `#[export]` fn return type of `Option<{integer}>` (reproduced with both `Option<u8>` and `Option<i32>`)

I could **not** reproduce it with any of the following (each compiles fine):
- a fieldless-only `#[data]` enum as the parameter (e.g. `enum Signal { A, B }`)
- `Option<{data enum}>` or `Option<String>` as the return type instead of `Option<{integer}>`
- a plain primitive parameter (no `#[data]` type involved) with `Option<{integer}>` return
- swapping the `Option<u8>`-returning function for one returning `bool`

So it's specifically the pairing of a data-carrying `#[data]` enum parameter with an `Option<{integer}>` return that breaks — not either condition alone, and the field syntax (named vs. tuple) of the variant doesn't change the outcome.

Levyks:
Fixes #621, The bug was found by an AI, as well as the suggested fix, but I've reviewed and it makes sense, as we're just replicating a pattern that's already used elsewhere (`DirectVec`), I've also manually validated that it fixes the issue I was having in my own crate.

## Summary

An `#[export]` function that takes a `#[data]` enum with a data-carrying
variant (named-field or tuple) as a parameter, and returns
`Option<{integer}>`, failed to compile:

```
error[E0308]: mismatched types
  --> src/lib.rs:22:1
   |
22 | #[export]
   | ^^^^^^^^^
   | expected `FfiBuf`, found `f64`
   | expected `boltffi::__private::FfiBuf` because of return type
```

## Root cause

`EncodedReturnStrategy::OptionScalar` in `boltffi_macros/src/exports/function.rs`
generates two functions — a wasm one returning `f64` (NaN-sentinel encoded)
and a native one returning `FfiBuf` — via `build_f64_wasm_return_exports`.

Both bodies were built from a single, shared `conversions`/`call_args` pair
computed once above the whole `match` (using the un-suffixed
`return_abi.invalid_arg_early_return_statement()`), rather than computed
per-platform. That shared conversion code's invalid-argument early return
resolves to the wasm convention (`return f64::NAN;`), so the *native*
function's parameter-decoding failure paths returned `f64::NAN` from a
function declared to return `FfiBuf` — a hard type error.

The neighboring `EncodedReturnStrategy::DirectVec` branch (same `match`, ~30
lines below) already does this correctly: it computes separate
`native_on_error`/`wasm_on_error` and separate `native_conversions`/
`wasm_conversions` via two `transform_params` calls, and uses each in its
respective body. This fix applies that exact pattern to `OptionScalar`.

Confirmed via `cargo expand` on a minimal repro — before the fix, the
generated native function's error paths read `return f64::NAN;`; after,
they correctly read `return ::boltffi::__private::FfiBuf::default();`.

## Fix

`boltffi_macros/src/exports/function.rs`: in the `OptionScalar` branch,
compute `native_conversions`/`native_call_args` and `wasm_conversions`/
`wasm_call_args` separately (mirroring `DirectVec`), instead of sharing one
`conversions`/`call_args` pair across both platform bodies.

## Regression test

Added `examples/option_scalar_data_param_regression/`, a minimal standalone
crate (excluded from the main workspace, like `examples/demo`) with:
- `Signal` — a `#[data]` enum with a tuple data-carrying variant
- `maybe_count(signal: Signal) -> Option<u8>` — the exact shape that failed
- A test calling the raw `extern "C"` symbol directly and checking both the
  `Some` and `None` cases cross correctly

This can't be added to `boltffi_tests` (the in-workspace test crate):
`boltffi_tests`'s `build.rs` unconditionally sets `BOLTFFI_BINDING_EXPANSION`,
which routes `#[export]`/`#[data]` through a separate, build-script-driven
IR expansion path (`__boltffi_ir_expansion`) that bypasses
`ffi_export_item_impl`/`build_f64_wasm_return_exports` entirely — so a test
added there would pass regardless of this bug. Verified this directly: the
same fixture added to `boltffi_tests` passed identically with the fix
stashed out. `examples/option_scalar_data_param_regression` uses a plain
`path` dependency on `boltffi`, like `examples/demo`, so it exercises the
same "ordinary consumer" codegen path the bug actually lives in.

## Verification

- New regression test fails to compile (same `E0308`/`FfiBuf`/`f64` error)
  on a clean build with the fix reverted; compiles and passes with the fix
  applied. Confirmed from a fully clean `target/` in both states.
- `cargo build --workspace`: clean.
- `cargo test -p boltffi -p boltffi_tests`: all passing (no regressions).


engali94:
nice catch @Levyks one thing let us reuse the demo examp,le for this instead of creating a new lib

Levyks:
done @engali94, also applied the same thing to my other PR #624

engali94:
@Levyks thanks! one more thing we need to call the new api from all demo cases to make the CI green and make sure the solution tested end to end

Levyks:
@codex review

engali94 (approved):
Thankss @Levyks

PR #626, opened 2026-07-07
Fix `r#type`-style raw identifier fields producing invalid encode buffer names

### LINKED ISSUE #625: `#[data]` struct field named with a raw identifier (`r#type`) panics WireEncode codegen with invalid identifier error
A `#[data]` struct with a field that requires a Rust raw identifier — e.g.
`r#type` (needed for any field literally named `type`, `fn`, `match`, etc.)
— panics at macro-expansion time.

## Reproduction

```rust
#[boltffi::data]
#[derive(Clone, Debug, PartialEq)]
pub struct TypedEvent {
    pub id: i64,
    pub r#type: Option<String>,
}
```

```
error: proc-macro derive panicked
message: `"__boltffi_buf_r#type"` is not a valid identifier
```

## Root cause

In `boltffi_macros/src/data/expansion/wire.rs`,
`StructWireExpansion::render_wire_encode_impl` builds each field's scratch
buffer identifier via:

```rust
let field_buffer =
    syn::Ident::new(&format!("__boltffi_buf_{}", field_name), field_name.span());
```

`format!`'s `Display` impl for a `syn::Ident` includes the literal `r#`
prefix, so for `field_name = r#type` this produces the string
`__boltffi_buf_r#type`, which is not a valid Rust identifier, and
`syn::Ident::new` panics.

## Workaround

Rename the field to avoid the raw identifier (e.g. `event_type` with
`#[serde(rename = "type")]` if the wire/JSON name must stay `type`).

## Fix

Build the identifier with `quote::format_ident!("__boltffi_buf_{}", field_name)`
instead — `format_ident!`'s `IdentFragment` handling unraws `r#`-prefixed
identifiers before splicing, which is exactly the pattern already used
correctly elsewhere in the same codegen.

Levyks:
Fixes #625 

## Problem

A `#[data]` struct field requiring a Rust raw identifier (e.g. `r#type`,
needed for any field literally named `type`, `fn`, `match`, etc.) broke
the generated `WireEncode::encode_to`, panicking at macro-expansion time
with `"__boltffi_buf_r#type" is not a valid identifier`.

Fixes #<issue-number>

## Root cause

`StructWireExpansion::render_wire_encode_impl` in
`boltffi_macros/src/data/expansion/wire.rs` built each field's per-field
scratch buffer identifier via:

```rust
syn::Ident::new(&format!("__boltffi_buf_{}", field_name), field_name.span())
```

`format!`'s `Display` impl for an `Ident` includes the literal `r#`
prefix, so a field named `r#type` produced the invalid identifier
`__boltffi_buf_r#type`.

## Fix

Switched to `quote::format_ident!("__boltffi_buf_{}", field_name)`, which
unraws `r#`-prefixed identifiers before splicing them into a new one —
the same helper already used correctly elsewhere in this codegen. Purely
a rename of the internal scratch-buffer identifier; no behavior change
for any field that isn't a raw identifier.

## Regression coverage

Added `TypedEvent`/`echo_typed_event` to
`examples/demo/src/records/keyword_fields.rs`, following the existing
`demo_case` convention used throughout that crate, with a field
literally named `type` to guard against this regressing.

## Verification

- `examples/demo` builds clean with the fix; reverting the one-line fix
  reproduces the exact panic and fails to compile, confirming the
  fixture catches the bug.


Levyks:
@codex review

Levyks:
@codex review

engali94:
The affected legacy scanner and macro expansion were removed during the IR migration so closing this OR

PR #628, opened 2026-07-07
Add InternedString support

ianks:
Target-agnostic `InternedString<P>`, a UTF-8 value that crosses the wire as either a static pool id or dynamic bytes. It's for open-but-mostly-known string sets, think user-agent browser/engine/OS names, where the common values are known ahead of time but you still need a fallback for anything you haven't seen.

Split out of the Ruby-target PR (#518) as its own base, mirroring the #593 (borrowed direct record params) split, since it's a general BoltFFI feature and not specific to Ruby.

### Usage

You declare the known set with a pool macro, then use `InternedString<Pool>` anywhere a string crosses the boundary:

```rust
boltffi::interned_string_pool! {
    pub BrowserName {
        Chrome = "Chrome",
        Safari = "Safari",
    }
}

// InternedString<BrowserName>::from("Chrome")  -> static id (1 tag byte + u32)
// InternedString<BrowserName>::from("Yandex")  -> dynamic bytes (fallback)
```

Known values encode as a small pool id; unknown values fall back to the normal length-prefixed string. Decode canonicalizes the other direction too, a dynamic payload that happens to name a pool value (`"Chrome"`) normalizes back to the static id so it compares equal to the generated constant.

### Wire format

```
u8 tag
tag = 0  => u32 pool id
tag = 1  => length-prefixed UTF-8 bytes (dynamic fallback)
```

### Capability gate

The tagged shape means a target has to actually understand it, a plain string decoder would misparse the leading tag byte. So `InternedString` is gated behind a `BindingCapability`: any declaration that references it requires the capability (detected by walking the decl's type refs), and a host that doesn't advertise support fails generation with a clear error instead of silently emitting a wrong decoder. No host advertises it on this branch; the Ruby target opts in via its capability set in #518.

### Layout

- `boltffi_core`: `InternedString<P>` + `InternedStringRepr {Interned(id), Dynamic}`, tagged wire encode/decode.
- `boltffi_macros`: `interned_string_pool!`.
- `boltffi_scan`: scans the pool macro (bare, `boltffi::`, and `boltffi_core::` qualified) and the type.
- `boltffi_binding`: IR + lowering, plus the capability requirement.
- `boltffi_backend`: the `InternedString` capability + gate.

Ruby's static-`VALUE`-table rendering is layered on top in #518. Standalone build + tests green (`interned_string.rs`: static pool, dynamic fallback, tagged wire round-trip, dynamic canonicalization).


boltffi_binding/src/lower/codecs.rs
```diff
@@ -20,7 +20,7 @@ pub fn node(
 ) -> Result<CodecNode, LowerError> {
     Ok(match type_expr {
         TypeExpr::Primitive(primitive) => CodecNode::Primitive(Primitive::from(*primitive)),
-        TypeExpr::String | TypeExpr::Str => CodecNode::String,
+        TypeExpr::String | TypeExpr::Str | TypeExpr::InternedString { .. } => CodecNode::String,
```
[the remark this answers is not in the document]
ianks: Thanks — this was accurate as a silent-misparse risk before the capability gate. I kept the codec node as `CodecNode::String` because #628 now requires `BindingCapability::InternedString` for any declaration referencing the type, and no non-tag-aware host advertises that capability. That makes non-Ruby generation fail with a clear `BindingCapability` error instead of emitting an incorrect plain-string decoder; Ruby opts in on the stacked Ruby PR.

boltffi_scan/src/marked.rs
```diff
@@ -237,6 +278,19 @@ fn custom_type_path(path: &syn::Path) -> bool {
...
 
+fn interned_string_pool_path(path: &syn::Path) -> bool {
+    let segments = path
+        .segments
+        .iter()
+        .map(|segment| segment.ident.to_string())
+        .collect::<Vec<_>>();
+    match segments.as_slice() {
+        [name] => path.leading_colon.is_none() && name == "interned_string_pool",
+        [namespace, name] => namespace == "boltffi" && name == "interned_string_pool",
```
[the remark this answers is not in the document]
ianks: Good catch — fixed in `7eab0156` by accepting `boltffi_core::interned_string_pool!` in the pool marker matcher, with a regression test covering the core-qualified macro path.

boltffi_binding/src/ir/closure.rs
```diff
@@ -118,6 +118,7 @@ impl fmt::Display for TypeSignature<'_> {
             TypeExpr::Enum { id, .. } => formatter.write_str(&source_type_signature(id.as_str())),
             TypeExpr::Class { id, .. } => formatter.write_str(&source_type_signature(id.as_str())),
             TypeExpr::Custom { id, .. } => formatter.write_str(&source_type_signature(id.as_str())),
+            TypeExpr::InternedString { .. } => formatter.write_str("InternedString"),
```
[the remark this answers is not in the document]
ianks: Good catch — fixed in `7eab0156` by including the recorded pool path in the inline closure type signature, so `Fn(InternedString<A>)` and `Fn(InternedString<B>)` no longer collide. Added a regression test that asserts the generated closure symbol parts differ.

boltffi_scan/src/type_expr.rs
```diff
@@ -467,6 +500,19 @@ fn ast_path_without_arguments(path: &syn::Path) -> Path {
...
 }
 
+fn interned_string_source_path(type_path: &syn::Path, pool_path: &syn::Path) -> Path {
+    let mut segments = path_segments_without_arguments(type_path).collect::<Vec<_>>();
+    if let Some(last) = segments.last_mut() {
+        *last = PathSegment::with_arguments(
+            last.name.clone(),
+            vec![GenericArgument::Const(ConstExpr::Raw(
+                pool_path.to_token_stream().to_string().replace(' ', ""),
+            ))],
```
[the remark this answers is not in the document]
ianks: Good catch — fixed in `7eab0156` by resolving the pool path through `DeclaredTypes` and recording it as a crate-qualified pool argument for the `InternedString<Pool>` source path. That avoids rendering root-level wrappers with a nested module's local `BrowserName` spelling.

boltffi_core/src/interned_string.rs
```diff
@@ -0,0 +1,140 @@
...
+                    Self {
+                        repr: InternedStringRepr::Interned(id),
+                        _pool: PhantomData,
+                    },
+                    1 + used,
+                ))
+            }
+            1 => {
+                let (value, used) = String::decode_from(&buf[1..])?;
+                Ok((Self::dynamic(value), 1 + used))
```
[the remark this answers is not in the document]
ianks: Good catch — fixed in `7eab0156` by canonicalizing tag=1 dynamic payloads through `InternedString::from_str`, so known dynamic strings normalize to the static pool representation and compare equal to generated constants. Added a regression test for decoding dynamic `Chrome` into `BrowserName::CHROME`.

boltffi_binding/src/lower/codecs.rs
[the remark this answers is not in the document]
engali94: I am afraid that once Ruby advertises support the shared read/write plan still says “normal string” but Ruby then has to ignore the codec plan and inspect `TypeRef` itself, that breaks the IR rules and will drift for functions, callbacks, collections, and other crossings

boltffi_backend/src/core/capabilities.rs
```diff
@@ -147,12 +157,23 @@ where
...
+    /// its type tree (fields, parameters, return types, stream items, constant
+    /// values, or custom-type representations).
     pub fn from_bindings<S: Surface>(bindings: &Bindings<S>) -> Self {
         bindings
             .decls()
             .iter()
             .fold(Self::new(), |requirements, decl| {
-                requirements.require(BindingCapability::from_decl(decl))
+                let requirements = requirements.require(BindingCapability::from_decl(decl));
+                if DeclarationRef::from(decl).contains_interned_string() {
```
engali94: i think partial generation can still get around this check so `from_bindings` is only used for complete generation but partial generation checks the declaration kind so if the target that supports functions but not InternedString can still try to render a function containing one it is better if we apply this requirement per declaration as well

boltffi_core/src/interned_string.rs
```diff
@@ -0,0 +1,140 @@
...
+    /// The caller must ensure `id` names a value in `P`'s generated pool.
+    pub const unsafe fn from_id_unchecked(id: u32) -> Self {
+        Self {
+            repr: InternedStringRepr::Interned(id),
+            _pool: PhantomData,
+        }
+    }
+
+    /// Builds a dynamic fallback value.
+    pub fn dynamic(value: impl Into<String>) -> Self {
```
engali94: seems here we can still create two values for the same string for example  `dynamic("Chrome")` will not equal the generated `CHROME` constant even though both represent "Chrome" string maybe this should not be public?

boltffi_macros/src/interned_string.rs
```diff
@@ -0,0 +1,81 @@
...
+        braced!(content in input);
+        let mut entries = Vec::new();
+        while !content.is_empty() {
+            let entry_name = content.parse()?;
+            content.parse::<Token![=]>()?;
+            let value = content.parse()?;
+            if content.peek(Token![,]) {
+                content.parse::<Token![,]>()?;
+            }
+            entries.push(PoolEntry {
```
engali94, line 33: should we reject duplicate values here? because for the current impl two entries can contain the same string but get different ids so they will compare as different values even though the actual string is the same

boltffi_scan/src/type_expr.rs
```diff
@@ -467,6 +502,51 @@ fn ast_path_without_arguments(path: &syn::Path) -> Path {
     )
 }
 
+fn interned_string_source_path(type_path: &syn::Path, pool_path: &Path) -> Path {
+    let mut segments = path_segments_without_arguments(type_path).collect::<Vec<_>>();
+    if let Some(last) = segments.last_mut() {
+        *last = PathSegment::with_arguments(
+            last.name.clone(),
+            vec![GenericArgument::Const(ConstExpr::Raw(path_spelling(
+                pool_path,
```
engali94: this is a type argument but we are storing it as raw const text maybe It worked because the later code prints the text back out but we lose the actual type information and need special cases elsewhere i think the pool should be stored as a proper typed path or maybe as a pool reference

boltffi_tests/tests/interned_string.rs
```diff
@@ -0,0 +1,78 @@
+use boltffi::__private::wire::{WireDecode, WireEncode};
+use boltffi::{InternedString, InternedStringPool, InternedStringRepr};
+
+boltffi::interned_string_pool! {
```
engali94, line 6: since we cant add this to the shared demo yet can we create a follow up issue for adding support to the other languages not to forget it?

boltffi_binding/src/ir/decl.rs
```diff
@@ -204,6 +204,25 @@ impl<'a, S: Surface> DeclarationRef<'a, S> {
         }
     }
 
+    /// Returns whether any type in this declaration is or contains an
+    /// [`InternedString`](crate::TypeRef::InternedString).
+    ///
+    /// Used by the capability gate: hosts that do not advertise the
+    /// `InternedString` capability will receive a clear error at generation
+    /// time instead of silently misparsing tagged bytes as plain strings.
+    pub fn contains_interned_string(self) -> bool {
```
engali94, line 235: we already have similar traversal for builtin codecs and other type checks so this will be easy to miss when a new shape is added. Can this capability come from the codec plans through the shared walker instead of adding it across the whole IR?

engali94 (approved):
Nice addition!
left few comments other than that all good

boltffi_backend/src/core/capabilities.rs
ianks: done

boltffi_core/src/interned_string.rs
ianks: done

boltffi_macros/src/interned_string.rs
ianks: done

boltffi_scan/src/type_expr.rs
ianks: done

boltffi_tests/tests/interned_string.rs
ianks: tracked in #667

boltffi_binding/src/ir/decl.rs
ianks: done

PR #639, opened 2026-07-08
Substitute generic type alias parameters during boltffi_scan resolution

### LINKED ISSUE #629: `boltffi_scan` doesn't substitute generic type aliases (`type Result<T> = ...`), breaking `generate python/kotlin/swift/kmp`
## Reproduction

```rust
#[boltffi::error]
#[derive(Debug, Clone, thiserror::Error)]
pub enum MyError {
    #[error("bad: {message}")]
    Bad { message: String },
}

pub type Result<T> = std::result::Result<T, MyError>;

#[boltffi::data]
#[derive(Debug, Clone, PartialEq)]
pub enum Wrapper {
    Signalled,
}

#[boltffi::export]
pub fn wrap() -> Result<Wrapper> {
    Ok(Wrapper::Signalled)
}
```

`boltffi generate python` (or kotlin/swift/kmp) fails:

```
error: BoltFFI metadata build scan failed: unsupported source type `Result<Wrapper>`
```

A single, non-generic `type Alias = ConcreteType;` already works fine
(covered by the existing `scans_custom_remote_used_through_type_alias`
test) — this is specifically about a *generic* alias, `type Alias<T> = ...`.

## Root cause

`boltffi_scan/src/declared_types.rs`:

- `TypeAlias` (used to represent a scanned `type X = ...;` declaration)
  only stores the alias's raw target type (`syn::Type`). `insert_alias`
  never records the alias's generic parameters (`alias.generics`), so
  there's no way to know that `T` in `std::result::Result<T, MyError>` is
  a placeholder that needs substituting.
- `resolve_custom_remote_with_aliases`, on finding a matching alias,
  recurses directly into `alias.target` — the raw, unsubstituted type.
  For `Result<Wrapper>`, it tries to resolve the literal (meaningless)
  `std::result::Result<T, MyError>` instead of the intended
  `std::result::Result<Wrapper, MyError>`.

This is invisible for non-generic aliases (nothing to substitute) and for
any usage where the type resolution never needs to look past the outer
alias name — which is why it's gone unnoticed until a crate actually
needs `Result<T>` with a locally-defined error type, a very ordinary
pattern for any crate exposing its own error type over FFI.

## Suggested fix

Record each alias's generic type parameter names at index time. At
resolution time, extract the use site's actual type arguments (from the
`syn::Path`'s `AngleBracketed` arguments) and substitute them into the
alias's target type tree before resolving further.

Levyks:
Fixes #629

## Problem

Crates that define a local generic `Result` alias — a very common pattern for exposing a crate-specific error type over FFI — fail metadata scanning used by `boltffi generate python/kotlin/swift/kmp`:

```rust
#[boltffi::error]
#[derive(Debug, Clone, thiserror::Error)]
pub enum MyError {
    #[error("bad: {message}")]
    Bad { message: String },
}

pub type Result<T> = std::result::Result<T, MyError>;

#[boltffi::data]
#[derive(Debug, Clone, PartialEq)]
pub enum Wrapper {
    Signalled,
}

#[boltffi::export]
pub fn wrap() -> Result<Wrapper> {
    Ok(Wrapper::Signalled)
}
```

Before this fix, generation fails with:

```
error: BoltFFI metadata build scan failed: unsupported source type `Result<Wrapper>`
```

Non-generic aliases (`type Alias = ConcreteType;`) already worked — covered by `scans_custom_remote_used_through_type_alias`. The gap was specifically **generic** aliases like `type Result<T> = ...`.

## Root cause

`boltffi_scan` indexed type aliases but never recorded their generic parameters, and never substituted the use site's actual type arguments before resolving further:

1. **`insert_alias`** stored only the raw target (`std::result::Result<T, MyError>`) with no record that `T` is a placeholder.
2. **`resolve_custom_remote_with_aliases`** recursed into that unsubstituted target when resolving custom remotes through aliases.
3. **`type_expr::path`** ran `standard_type` on the alias use site (`Result<Wrapper>`, one type argument) instead of the expanded type (`std::result::Result<Wrapper, MyError>`, two arguments), so `Result` detection rejected it.
4. **Cross-module usage** rescanned the substituted target in the caller's scope, so relative names in the alias definition (e.g. `MyError` in `errors::Result<T>`) were resolved in the importing module instead of where the alias was declared.

## Fix

- Record each alias's type parameters (including defaults) at index time.
- At resolution time, extract the use site's type arguments and substitute them into the alias target before recursing.
- Scan use-site type arguments in the caller's scope, but scan the alias's written target in the alias's **definition** scope.
- In `type_expr::path`, expand aliases **before** `standard_type` so `Vec`/`Option`/`Result` detection sees the substituted arity.
- When a use site omits a trailing generic parameter that has a default (`type Result<T, E = MyError> = ...`), fall back to that default.

## Verification

- `scans_function_return_through_a_generic_type_alias` — same-module `Result<Point>` and `Result<u32>` resolve correctly.
- `scans_generic_type_alias_used_from_a_module_that_never_imported_its_error_type` — `errors::Result<T>` used from `api` resolves `MyError` in `errors`, not `api`.
- `scans_generic_type_alias_with_a_defaulted_parameter_omitted_at_the_use_site` — `Result<Point>` with `type Result<T, E = MyError> = ...` picks up the default error type.
- `cargo test -p boltffi_scan` — 240 passing (237 existing + 3 new), no regressions.


engali94:
@Levyks Nice work coudl you fix the lint errors

Levyks:
i'm not sure if this should still be merged considering the discussion in #665 , but anyways, I fixed the lint errors

engali94:
closing will be resolved as part of https://github.com/boltffi/boltffi/issues/665

PR #647, opened 2026-07-09
fix: lower Result<Class, E> returns as object handle, not wire-encoded record

Marcuss2:
## Problem

A return type of `Result<Class, Error>` (or `Result<Self, Error>`) on an instance method, static method, or free function was misclassified by the legacy macro lowering path as a wire-encoded buffer. The generated code emitted `FfiBuf::wire_encode(&result)` on the *whole* `Result<Class, Error>`, which requires `Class: WireEncode`/`WireDecode`. A class is an object handle, not a wire value — so this produced a trait-bound compile error:

> `the trait bound \`Class: WireDecode\` is not satisfied`

This is the "treats Class as a record, not a class reference" symptom. Factory constructors (`Result<Self, E>`) already had a special-case bypass, but regular returns did not.

## Root cause

`ReturnLoweringContext::object_handle_return` only recognized bare `Class` and `Option<Class>`, not `Result<Class, E>` / `Result<Self, E>`. So `classify_value_return_strategy` fell through to the `StandardContainer::Result` branch and picked `WireEncoded`.

## Fix (`boltffi_macros/src/lowering/returns/`)

- **`model.rs` — `object_handle_return`**: now matches `StandardContainer::Result { ok, .. }` and, when `ok` is a class/Self pointee, returns `ObjectHandleReturn::fallible(...)`. When `ok` is a record/data type, `object_handle_pointee` returns `None`, so `Result<Record, E>` still falls through to wire encoding — the correct path, since records implement `WireEncode`/`WireDecode`.
- **`model.rs` — `ObjectHandleReturn`**: gained a `fallible: bool` field (`is_fallible()` accessor). `lower_type` sets `ErrorReturnStrategy::StatusCode` when fallible.
- **`lower.rs` — `raw_pointer_expression`**: when `is_fallible()`, emits `match { Ok(value) => Box::into_raw(...), Err(error) => { set_last_error(...); null_mut() } }` — mirroring the existing factory-constructor fallible path. Instance methods, static methods, and free functions all route object-handle returns through this one helper, so all three call sites are fixed.

`classify_value_return_strategy` needed no change: `object_handle_return(rust_type)` is checked first, so `Result<Class, E>` now resolves to `ObjectHandle` before reaching the `StandardContainer::Result` branch.

### Unit tests (`boltffi_macros/src/exports/methods.rs`)
4 new tests:
- `instance_method_returning_result_of_exported_class_does_not_wire_encode` — **reproduces the bug** (fails before the fix, passes after): asserts no `wire_encode`, Ok payload becomes `Box::into_raw(Box::new(value))`, Err uses `set_last_error`.
- `instance_method_returning_result_of_self_does_not_wire_encode` — `Result<Self, E>` on an instance method (non-constructor).
- `static_method_returning_result_of_exported_class_does_not_wire_encode` — static `Result<OtherClass, E>`.
- `instance_method_returning_result_of_record_still_wire_encodes` — **regression guard**: `Result<UserProfile, String>` (a `#[data]` record, not a class) must still wire-encode the whole Result and must NOT lower to an object handle or use `set_last_error`.

### Integration tests (`boltffi_tests/`)
- Added `FallibleService::try_make_counter(&self, i32) -> Result<TestCounter, FixtureError>` — a class ok + `#[error]` enum err, matching the `Result<Image, DemoError>` shape from the bug report.
- C-harness call in `build.rs` exercising the success path.
- Two FFI-level tests in `class_ffi.rs`:
  - `try_make_counter_positive_returns_class_handle`: success returns a live `TestCounter` handle whose `get()` echoes the initial value (7).
  - `try_make_counter_negative_returns_invalid_input_error`: negative initial returns `FixtureError::InvalidInput`.

## Verification

- `boltffi_macros`: 263 passed (2 pre-existing wasm32-target-not-installed failures, unrelated).
- `boltffi_tests`: 258 passed (including existing `Result<Record, E>` tests `try_rect`/`try_message` in `crossing_behavior.rs`, confirming no regression).
- `boltffi_binding` (327) and `boltffi_core` (218): no regressions.
- `just fmt-check` and `just lint` (workspace clippy `-D warnings`): clean.

### DIFF
--- a/boltffi_macros/src/exports/methods.rs
+++ b/boltffi_macros/src/exports/methods.rs
@@ -1911,6 +1911,224 @@ mod tests {
     }
 
     #[test]
+    fn instance_method_returning_result_of_exported_class_does_not_wire_encode() {
+        // Regression: `Result<Class, E>` returns must not lower to a
+        // whole-Result `wire_encode`, which would require the class type
+        // to implement `WireEncode`/`WireDecode`. The class is a handle,
+        // not a wire value, so the success payload must cross as an
+        // object handle and the error must cross via the last-error
+        // channel — mirroring the factory-constructor fallible path.
+        let impl_block = parse_impl(
+            r#"
+            impl Map {
+                pub fn try_make_marker(&self, value: i32) -> Result<Marker, String> {
+                    if value < 0 {
+                        Err("negative".to_string())
+                    } else {
+                        Ok(Marker)
+                    }
+                }
+            }
+            "#,
+        );
+        let method = impl_block
+            .items
+            .iter()
+            .find_map(|item| match item {
+                syn::ImplItem::Fn(method) => Some(method),
+                _ => None,
+            })
+            .expect("instance method should exist");
+        let type_name = impl_type_name(&impl_block).expect("impl type name should resolve");
+
+        let generated = generate_sync_method_export(
+            MethodCallable::new(method),
+            &type_name,
+            "Map",
+            &class_return_lowering(),
+            callback_registry(),
+        )
+        .expect("instance export should be generated")
+        .to_string();
+
+        // The bug: the whole Result was wire-encoded, emitting
+        // `FfiBuf :: wire_encode (& result)` which fails to compile
+        // because `Marker` (a class) does not implement WireEncode /
+        // WireDecode.
+        assert!(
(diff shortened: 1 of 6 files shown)


engali94 (approved):
Nice catch @Marcuss2 thanks!

PR #652, opened 2026-07-10
fix cfg attr scanning closes #630

### LINKED ISSUE #630: `#[cfg_attr(feature = "...", boltffi::data)]` is silently treated as unmarked, breaking `generate python/kotlin/swift/kmp`
## Reproduction

```rust
#[cfg_attr(feature = "boltffi", boltffi::data)]
#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
pub enum Signal {
    Ping,
    Pong,
}

#[cfg_attr(feature = "boltffi", boltffi::export)]
pub fn signal_name(signal: Signal) -> String {
    format!("{signal:?}")
}
```

built with `--features boltffi` (or as the crate's default feature).
`boltffi generate python` (or kotlin/swift/kmp) fails — not with a clear
error at `Signal`'s declaration, but with a confusing "unsupported source
type" pointing at some *other*, unrelated item elsewhere in the crate
(whichever macro invocation happens to trigger the crate-wide scan).

The exact same code with bare `#[boltffi::data]` / `#[boltffi::export]`
(no `cfg_attr`) works fine.

## Root cause

`boltffi_scan` parses source text directly via `syn` rather than going
through the compiler's macro-expansion pipeline, so it never sees
`cfg_attr` get expanded the way rustc would. `Marker::from_attribute`
(`boltffi_scan/src/marker.rs`) only matches a bare `boltffi::data` /
`boltffi::export` / `boltffi::error` / `boltffi::skip` path. A
`cfg_attr(...)` attribute's own path is `cfg_attr`, not `boltffi::data`,
so `marker_name`/`from_attribute` never match it — the item is silently
treated as unmarked, exactly as if it had no marker attribute at all.

This specifically blocks the common pattern of gating every FFI
annotation behind a Cargo feature so a crate can also build with zero FFI
dependency (`default-features = false`) — `cfg_attr` is the only way to
make an attribute macro invocation itself conditional in stable Rust.

## Suggested fix

Add a `cfg_attr` expansion step (evaluating the wrapped predicate against
the crate's real active Cargo features, e.g. via `CARGO_FEATURE_*` env
vars already read elsewhere in this crate for module-level `#[cfg(...)]`
gating) that runs before marker detection, replacing
`cfg_attr(predicate, wrapped...)` with `wrapped...` when `predicate` is
active and dropping it otherwise — mirroring what the compiler's own
expansion does.

engali94:
closes #630


osa8bit (approved):

engali94:
that would be fixed once #685 lands so closing

PR #654, opened 2026-07-10
Migrate C# to the new IR backend

### LINKED ISSUE #653: Migrate C# to new model
This issue migrates C# to the new model the other languages have migrated to

mhedgpeth:
## Summary

- route C# generation through the new backend IR model by default
- migrate values, errors, handles, async calls, callbacks, closures, streams, and constants
- remove the legacy C# renderer and its obsolete snapshots

## Testing

- `cargo test -p boltffi_backend target::csharp --offline`
- `cargo test -p boltffi_cli commands::generate::tests::csharp_generate_uses_ir_route_without_ir_flag --offline`
- `cargo clippy -p boltffi_backend -p boltffi_bindgen -p boltffi_cli --offline -- -D warnings`
- generated the full demo through the default C# route with no unsupported diagnostics
- built the generated .NET 8 project with zero warnings and zero errors

## Notes

- The full workspace test run reaches an environment-specific MSVC C11 atomics failure in `boltffi_tests`.
- The broader CLI suite has existing Windows path and tool-resolution failures unrelated to this migration.

Closes #653


mhedgpeth:
This PR looks ready, and includes #662 fix in it (and the additional demo tests). I also compared it to the recent java one. I'm doing a final step by step review to make sure everything fits and I can completely own this going forward.

mhedgpeth:
I did the review of this and am really excited about this direction in the project, now that I am up to speed with it. The centralization of decision making is really good!

In this change I focused on making C# standard with the others, and in doing so abandoned my C# AST rule that strings must be represented in either C# AST that you could read in a C# book, or in templates. This enables code that I don't really like, like in `./boltffi_backend/src/target/csharp/render/mod.rs:1613`. But rather than try to correct that here, I'll take that on separately because the rule should probably be consistent in all the backends; in my mind consistency is better than my own sense of comfort here. So I'll tackle that separately.

engali94 (approved):
🎉 NICE!

PR #657, opened 2026-07-13
Emit fun interface for single-method Kotlin callbacks

akesson:
## Summary

A capability trait with a single method currently renders as a plain Kotlin `interface`, which Kotlin cannot SAM-convert — every implementation site has to spell out an `object :` expression:

```kotlin
// before
store.install(object : Listener {
    override fun onValue(value: UInt): UInt = value + 1u
})

// after
store.install(Listener { value -> value + 1u })
```

This PR emits `fun interface` when a callback has exactly one abstract method. Multi-method callbacks keep rendering as plain interfaces.

## Fix

- `boltffi_backend/templates/target/kotlin/callback.kt`: the interface header becomes `{% if callback.fun_interface() %}fun {% endif %}interface …`.
- `boltffi_backend/src/target/kotlin/render/callback.rs`: adds the `fun_interface()` predicate on `Callback` — true when `methods.len() == 1`, mirroring the language rule (Kotlin permits the `fun` modifier only on interfaces with exactly one abstract method).

The closure template in the same directory already emits `fun interface` for the identical shape (`templates/target/kotlin/closure.kt`); this applies that existing pattern to callbacks.

## Suspend callbacks are included

The predicate is "exactly one abstract method", not "one *sync* method": a single-method async callback renders as `fun interface Listener { suspend fun load(key: UInt): String }`. Suspend members in `fun` interfaces are legal since Kotlin 1.6, well below the versions this repo pins (Kotlin example: 2.0.21; KMP template: 2.3.21), and suspend callbacks are arguably where lambdas help most. Verified by compiling the rendered snapshot output (see below).

Adding `fun` is source- and binary-compatible for existing implementers (`object :` expressions and implementing classes are unaffected). If a callback later grows a second method, the modifier comes off again — lambda call sites then break, but any second abstract method already breaks `object :` implementers the same way, so this does not add a new evolution hazard.

---

Authored with AI assistance (Claude Fable); reviewed and validated manually.

### DIFF
--- a/boltffi_backend/src/target/kotlin/render/callback.rs
+++ b/boltffi_backend/src/target/kotlin/render/callback.rs
@@ -268,6 +268,13 @@ impl Callback {
         &self.methods
     }
 
+    /// Kotlin permits the `fun` modifier only on interfaces with exactly one
+    /// abstract method; a `fun interface` gains SAM conversion, so callers can
+    /// pass a lambda instead of an `object :` expression.
+    pub fn fun_interface(&self) -> bool {
+        self.methods.len() == 1
+    }
+
     pub fn handle_methods(&self) -> &[HandleMethod] {
         &self.handle_methods
     }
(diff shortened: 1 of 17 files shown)


engali94 (approved):
Nice Catch!
Thanks!

PR #658, opened 2026-07-13
fix wire decode field name collision with internal bindings

Marcuss2:
## Summary
The #[data] macro generates wire encode/decode implementations for structs and enums. In the decode path, each field was destructured as:

    let (#field_name, size) = #decode_expr?;
    position += size;

If a struct or enum variant had a field named `size`, this produced `let (size, size) = ...` — a duplicate binding compile error. A field named `position` would silently shadow the position tracker, corrupting all subsequent field decodes.

The same class of bug existed in the enum encode path where `written` (the byte counter) could collide with a variant field named `written`.

## Fix (boltffi_macros/src/data/expansion/wire.rs):
- Rename `size` → `__boltffi_size` in all generated decode code (struct, enum unnamed, enum named).
- Rename `position` → `__boltffi_position` in all generated decode code (struct, enum unnamed, enum named).
- Rename `written` → `__boltffi_written` in all generated enum encode code (unnamed, named).
- The `__boltffi_` prefix matches the existing convention used throughout the generated code.

## Tests:
- boltffi_macros: 6 unit tests verifying the generated token stream uses prefixed bindings and does not produce duplicate patterns for fields named `size`, `position`, or `written`.
- boltffi_core: 4 integration tests in wire_records.rs doing real wire round-trips with structs and enums that have fields named `size` and `position`.

### DIFF
--- a/boltffi_core/tests/wire_records.rs
+++ b/boltffi_core/tests/wire_records.rs
@@ -1029,6 +1029,58 @@ mod records {
         let (decoded, _) = ManyFields::decode_from(&buf).unwrap();
         assert_eq!(decoded, original);
     }
+
+    #[test]
+    fn record_with_size_field_roundtrip() {
+        // A field named `size` used to collide with the internal `size`
+        // binding in the generated wire decode code, producing
+        // `let (size, size) = ...` — a duplicate binding error.
+        #[data]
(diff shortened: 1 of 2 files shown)


engali94 (approved):
Good catch!
thank you!

PR #662, opened 2026-07-14
Fix C# I1 return marshalling leaking onto FfiBuf returns

akesson:
Fallible bool functions (`Result<bool, E>`) cross the FFI boundary as an `FfiBuf` wire envelope, but the C# renderer still emitted `[return: MarshalAs(UnmanagedType.I1)]` on the DllImport — that directive is only valid on primitive bool returns, so calling any such function threw `MarshalDirectiveException` at runtime.

**Changes**
- Guard the `I1` directive on whether the return actually crosses as a primitive bool, hoisted into `CSharpCallablePlan::native_return_needs_i1_marshalling()` and used by all 8 DllImport sites in `native.txt` (4 sync sites had no guard; 4 async sites already did).
- Regression tests in `emit.rs` covering both function and class-method fallible bool returns; the existing test asserting `I1` is still emitted for plain bool returns covers the other polarity.
- Demo coverage: `is_even(value: i32) -> Result<bool, String>` in `examples/demo/src/results/basic.rs`.

---

Authored with AI assistance (Claude Fable); reviewed and validated manually.


engali94:
Nice catch @akesson 
one thing currently C# being migrated to new IR based backend in #654 and the old one being removed, I am afraid your fix might not land there if you can stack it on top of that pr would be great!
cc: @mhedgpeth

akesson:
@engali94 
Thanks — I checked, and from what I can see, stacking isn't needed: the bug doesn't exist in the new IR backend. There, `return_marshal_i1` is derived alongside the native return type, so `Result<bool, String>` renders as `FfiBuf NativeIsEven(int value, [MarshalAs(I1)] out bool boltffiResult)` — no `[return: MarshalAs(I1)]`. I generated the demo through #654 and called it from .NET to confirm; it passes.

mhedgpeth:
Thanks for this @akesson I'll make sure this is rightly supported by the time I get to the end of it, hopefully by the end of the week.

mhedgpeth:
I confirmed this is working on #654 and pulled in the new demo tests! :)

Thanks again @akesson I'm going to close this but we'll get it in the upcoming merge.

Issue #665, opened 2026-07-14
RFC: per-invocation metadata capture — retire the source re-scan inside the metadata build


akesson:
## Summary

The experimental metadata path (#402, closed) moved binding discovery to the compiled artifact: macros emit a link-section envelope, bindgen reads it back (#538, #539, #540, #549). But the *content* of that envelope still comes from a whole-crate **source re-scan** running inside the first macro invocation (`metadata_build.rs`, single-shot guard from #555). So the new path inherits every blind spot of reading source text off disk, and a family of bugs keeps growing from that one root.

This RFC proposes completing the #402 direction: each macro invocation emits its **own** metadata record, and type references are resolved **by rustc** instead of by `boltffi_scan`. I have a working prototype on stable 1.95 that de-risks the hard parts, and one breaking API decision that needs a maintainer call before any implementation work makes sense.

The contributor docs already state the destination: *"the new path is `scan -> ast -> bindings -> backend`, with metadata coming from the compiled artifact."* This proposal makes that literally true — the metadata would come from what the compiler compiled, not from a re-parse of the source.

## The recurring bug class

These are all symptoms of discovery-by-source-scan, each currently being patched case-by-case in the scanner:

| issue | symptom | root cause in the scan |
|---|---|---|
| #630 (open, fix #652), #618 (closed) | cfg-gated items silently unmarked or skipped | `ActiveCfg` approximates rustc's cfg evaluation |
| #629 (open, fix #639) | generic type aliases not substituted, breaking 4 generators | scanner reimplements name resolution |
| #80 (open) | scanner chokes on syntax outside exported items | whole-crate parse sees code rustc would cfg away |
| (tracked by this RFC) | `#[data]`/`#[export]` emitted by a macro or `include!`d from `OUT_DIR` silently absent from bindings, `boltffi generate` exits 0 | scanner never sees macro-expanded code |

The last row is the worst failure mode because it is silent — rustc happily compiles items that never reach the bindings. It has no separate issue; this RFC tracks it, and the prototype (#685) reproduces all three placements as passing tests on the new path.

Per-invocation capture eliminates the class, not the instances: cfg is evaluated by rustc (an inactive item never runs its macro, so it never emits a record), aliases are resolved by rustc, and macro-emitted items are captured because the emitted tokens *are* the discovery.

## Proposed mechanics

Three changes, all inside the experimental namespace, legacy scan untouched until promotion:

[...]

PR #685, opened 2026-07-16
RFC #665 prototype: per-invocation metadata and expansion capture

akesson:
Draft for inspection, per #665 — **not intended to merge as-is**. A workspace-excluded prototype (`prototypes/per_invocation_metadata/`; no production code touched):

- **metadata.** Every `#[data]` invocation emits its own const-built link-section record, with type references resolved by rustc through `TypeInfo<Tag>` consts instead of today's source scan. Macro-emitted and `include!(OUT_DIR)` items are captured; aliases, globs, re-exports, and cross-crate references resolve; cfg is evaluated for real; records survive `release-lto`; an unresolvable reference is a compile error at the exact field.
- **expansion.** Every `#[export]` invocation emits its own `extern "C"` wrapper, ABI projected through `Codec<Tag>::FfiType`, plus a record carrying the wrapper's link symbol. The acceptance test dlopens the cdylib and calls wrappers through symbol names read from their own records — including the two previously-invisible cases (`macro_rules!`-emitted and `include!(OUT_DIR)` exports). Each of the expander's four whole-crate dependencies has a proven per-invocation replacement (FINDINGS §6).
- **edges.** Every wrapper catches panics into a trailing call-status out-param, with the poison return supplied by the codec. Callbacks (repr(C) vtables pinned by the record), streams (subscribe/pop/free), and `#[export] impl` class methods (u64 handles) all expand per-invocation too. A wasm32 smoke test pins the two deviations: wasm-ld drops a dependency's unreferenced records, and `include!(OUT_DIR)` symbols hash per target dir.

[`FINDINGS.md`](https://github.com/boltffi/boltffi/blob/prototype/per-invocation-metadata/prototypes/per_invocation_metadata/FINDINGS.md) is the write-up: results, costs (the mandatory crate-root anchor accepted in #665, per-crate `custom_type!`, symbol-hash caveats), and the open edges (bindgen-side aggregation, async and the stream wake half, Rust-to-foreign callbacks, `custom_type!` values).


engali94:
@akesson This looks great please go ahead. I’d prefer N smaller PRs instead of one large PR.

Lets create a dedicated integration branch for this work and merge the PRs into it sequentially. That way, we can review and test each part independently without landing a partially completed architecture on main.

Once the entire stack is complete, integrated, and thoroughly tested, we can open one final PR from the integration branch into main and roll it out as the production implementation.

Also im fine with the mandatory crate root `boltffi::scaffolding!()` and per crate `custom_type!` registrations. 
we can keep #685 as the prototype and reference implementation I don’t think further prototype work is necessary unless the production implementation exposes another design question.

PR #690, opened 2026-07-17
fix(java): emit JNI callback helpers for registered callback types

### LINKED ISSUE #688: pack java generates incomplete jni_glue.c for crates that only take host callbacks as parameters
## Summary

For thin crates that register a callback trait and pass it into a constructor (but do not nest/return callbacks), generated `jni_glue.c` calls `boltffi_jni_callback_parameter` without including the `callback.c` helpers that define it. JNI compilation fails on every platform.

## Repro

Minimal export surface:

```rust
#[export]
pub trait MathExecutor: Send + Sync {
    fn execute(&self, value: i64) -> i64;
}

#[export]
impl MathWrapper {
    pub fn new(executor: Box<dyn MathExecutor>) -> Self { /* ... */ }
    pub fn execute(&self, initial_value: i64, iterations: u32) -> i64 { /* ... */ }
}
```

Then `boltffi pack java` / compile `jni_glue.c`.

## Observed

Native methods call:

```c
boltffi_jni_callback_parameter(..., boltffi_create_callback_...)
```

but the helper from `bridge/jni/callback.c` is not emitted. MSVC fails converting an assumed `int` return to `BoltFFICallbackHandle`; clang typically fails with an undeclared function.

## Cause

JNI source feature selection set `uses_callback_handles` only for nested/returned/handle-method callback shapes, not for “any registered callback type.” Constructors that *take* a host callback still need those helpers.

Demo crates usually exercise richer callback surfaces, so the fragment is pulled in and the bug stays latent.

## Scope

All OS targets that generate JNI glue (Java and Kotlin). Not Windows-specific.

## Expected

Generated `jni_glue.c` includes callback handle helpers whenever any callback type is registered, so thin callback-parameter crates pack cleanly everywhere.

Levyks:
Fixes #688 

## Summary

Emit `callback.c` handle helpers (and the exception helpers they use) whenever any callback type is registered.

Native methods that take a host callback call `boltffi_jni_callback_parameter`. Feature selection previously only pulled that fragment in for nested/returned/handle-method shapes, so thin crates that only pass a callback into a constructor generated incomplete `jni_glue.c` on every OS.

## Changes

- `boltffi_backend/src/bridge/jni/template/source/features/source.rs` — `uses_callback_handles` and `uses_exceptions` include `callbacks.has_registrations`

## Test plan

- [x] Minimal callback-parameter crate (`MathExecutor` + `MathWrapper`) packs/compiles JNI glue

## Notes

Cross-platform JNI codegen fix (Java and Kotlin). Independent of the MSVC C11 atomics change.


engali94 (approved):

engali94:
@Levyks all good thanks can you make sure that the ci is happy

engali94:
already fixed in #696.

PR #720, opened 2026-07-27
Add lock free async callback polling

engali94:
This changes async callback trait methods implemented by foreign code and called from Rust. It does not change exported Rust async functions.

 For a callback such as:

 ```rust
   #[async_trait::async_trait]
   pub trait DataSource {
       async fn load(&self, key: String) -> Result<Vec<u8>, String>;
   }
```

The value directions are:

 1. Rust calls the foreign implementation and sends parameters OutOfRust.
 2. The foreign implementation completes asynchronously.
 3. The completion payload or error returns IntoRust.
 4. The waiting Rust future resumes.

#### Native callbacks

 Native callbacks invoke the foreign implementation through its registered vtable. Previously, every generated async callback method emitted its own Arc<Mutex<...>>, completion function, waker storage, and poll_fn loop. Encoded payloads were also copied  from `FfiBuf `into a `Vec<u8>` before the future could consume them.

The generated path now uses ForeignCall:

``` rust
   let (status, result) = unsafe {
       ForeignCall::<FfiBuf>::start(|completion, completion_data| unsafe {
           ((*self.vtable).load)(
               self.handle,
               key_ptr,
               key_len,
               completion,
               completion_data,
           );
       })
   }
   .await;

   if status.is_err() {
       Err(decode_error(result))
   } else {
       Ok(decode_value(result))
   }
```

ForeignCall owns the completion payload, coordinates the callback and waiting task without a mutex, and safely handles synchronous, cross-thread, and late completion.

#### Wasm callbacks

Wasm callbacks invoke the host implementation through an imported function. Completion arrives later through the exported request-completion function. The generated path now creates an AsyncCallbackWait before making the import call:

``` rust
   let registry = AsyncCallbackRegistry::current();
   let request = registry.allocate();
   let wait = registry.wait(request);

   unsafe {
       imported_load(self.handle, request.as_u32(), key_ptr, key_len);
   }

   let completion = wait.await;

   if completion.code.is_success() {
       Ok(decode_value(&completion.data))
   } else {
       Err(decode_error(&completion.data))
   }
```

Dropping AsyncCallbackWait removes the pending request, so cancellation remains tied to the future’s lifetime.

### DIFF
--- a/boltffi_core/src/runtime/async_callback.rs
+++ b/boltffi_core/src/runtime/async_callback.rs
@@ -1,6 +1,8 @@
 use std::cell::RefCell;
 use std::collections::HashMap;
-use std::task::Waker;
+use std::future::Future;
+use std::pin::Pin;
+use std::task::{Context, Poll, Waker};
 
 /// Callback ABI for async completions that carry a result payload.
 ///
@@ -219,6 +221,13 @@ impl AsyncCallbackRegistry {
         self.with_state(|registry_state| registry_state.cancel(request_id))
     }
 
+    pub fn wait(&self, request_id: AsyncCallbackRequestId) -> AsyncCallbackWait {
+        AsyncCallbackWait {
+            request_id,
+            guard: Some(AsyncCallbackRequestGuard::new(request_id)),
+        }
+    }
+
     #[cfg(target_arch = "wasm32")]
     pub unsafe fn complete_from_ffi(
         &self,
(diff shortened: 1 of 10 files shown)


osa8bit (approved):

boltffi_core/src/runtime/async_callback.rs
```diff
@@ -271,3 +280,105 @@ impl Drop for AsyncCallbackRequestGuard {
...
+}
+
+impl Future for AsyncCallbackWait {
+    type Output = AsyncCallbackCompletion;
+
+    fn poll(self: Pin<&mut Self>, task: &mut Context<'_>) -> Poll<Self::Output> {
+        let wait = self.get_mut();
+        let registry = AsyncCallbackRegistry::current();
+        registry.set_waker(wait.request_id, task.waker().clone());
+        match registry.take_completion(wait.request_id) {
```
nxmxsxs, line 296: I believe this should be in the None branch below

boltffi_core/src/runtime/async_callback.rs
```diff
@@ -271,3 +280,105 @@ impl Drop for AsyncCallbackRequestGuard {
...
+    guard: Option<AsyncCallbackRequestGuard>,
+}
+
+impl Future for AsyncCallbackWait {
+    type Output = AsyncCallbackCompletion;
+
+    fn poll(self: Pin<&mut Self>, task: &mut Context<'_>) -> Poll<Self::Output> {
+        let wait = self.get_mut();
+        let registry = AsyncCallbackRegistry::current();
+        registry.set_waker(wait.request_id, task.waker().clone());
```
nxmxsxs, line 295: I believe this should be in the None branch below?

nxmxsxs (approved):

akesson:
The new `ForeignCall` state machine is exactly what Miri is for, but the CI Miri job only runs `-p boltffi -p boltffi_tests`, so the `foreign_call` unit tests (including the cross-thread completion one) never run under it. Ran `cargo +nightly miri test -p boltffi_core foreign_call` locally — all 7 pass clean. How about adding `-p boltffi_core` to the Miri step in `ci.yml` so that coverage sticks?

akesson (approved):

PR #724, opened 2026-07-27
improve wasm performance for strings and optional enums

jlucaso1:
Two independent wins on the WASM/TypeScript surface: optional C-style enums stop round-tripping through a wire buffer, and strings stop being encoded twice. Both were found while running `benchmarks/harnesses/wasm-bench`, which also turned out to be measuring itself — that fix is included and explained below, since it changes the numbers this PR should be judged against.

## Optional C-style enums

A C-style enum already crosses as its discriminant when returned bare, which is why `echo_direction` costs ~9ns. Wrapping it in `Option` fell back to the wire buffer:

```rust
pub fn find_direction(id: u32) -> Option<Direction>
```

```ts
// before
export function findDirection(id: number): Direction | null {
  const __boltffiReader = _module.takePackedBuffer(_exports.…find_direction(id));
  return __boltffiReader.readOptional(() => DirectionCodec.decode(__boltffiReader));
}
```

That is a wasm-side malloc, a buffer copy into a fresh `ArrayBuffer`, a `DataView`, a closure per call, and a free — for a value that fits in an `i32`. `Option<primitive>` has ridden the f64 scalar slot for a while; `classify_value_return_strategy` simply never considered an enum for it.

```ts
// after
export function findDirection(id: number): Direction | null {
  return _module.unpackOptionI32(_exports.…find_direction(id)) as Direction | null;
}
```

The cast is what keeps the declared type `Direction | null` rather than widening it to `number | null`.

**Scope.** Three gates keep this from reaching anything it was not measured on:

- **Returns produced by Rust only.** Params and callback returns keep the wire encoding, because their Rust side still decodes a buffer. The specialization is gated on the return direction being `OutOfRust`, so the two sides cannot disagree on the ABI — without that gate, a callback returning `Option<Direction>` would have had TypeScript packing a scalar while Rust still read a buffer.
- **wasm only.** `OptionScalar`'s native body is `FfiBuf::wire_encode` — byte-for-byte what `WireEncoded` already emitted — so Swift, Kotlin, Java, C# and Python are untouched by construction, not by a flag.
- **Reprs that survive the carrier only.** `i64`/`u64` stay wire-encoded, as do `isize`/`usize` (target-dependent width) and any repr outside the known set. An `i64` discriminant would come back rounded through `f64`.

## String encoding

`utf8ByteCount` was:

```ts
return UTF8_ENCODER.encode(value).length;
```

It encoded the whole string to read its length and discarded the result. `writeString` then encoded the same value again, so every string crossing the boundary was encoded twice, allocating a throwaway `Uint8Array` each time. A record with three strings cost ~4.4us to encode, dominated by the allocations.

Now the byte count is a scan that allocates nothing, and `writeString` uses `encodeInto` so the payload lands directly in the target buffer — ~119ns for the same record. The counts match `TextEncoder` exactly, including the 3-byte U+FFFD that an unpaired surrogate collapses to.

One wrinkle worth flagging: `encodeInto` can stop short on an exactly sized buffer for some non-ASCII inputs. `"ção".repeat(2000)` writes 9999 of 10000 bytes and reads 5999 of 6000 code units; one spare byte avoids it. Padding the buffer is not an option because `withWasmRegion` writers cannot grow, so a short write falls back to the allocating path. There is a regression test for that exact string, in a fixed-size wasm region.

## Benchmark harness

The harness reached the BoltFFI exports through a `Proxy` whose `get` trap ran `Reflect.has` + `Reflect.get` on every call, while wasm-bindgen was a plain CommonJS object — only one side paid. A `get` trap over a module namespace costs ~200-300ns per access, so the scalar benchmarks were measuring the trap: `noop` reported 101ns against wasm-bindgen's 3ns, and `simple_enum` 72x slower.

Flattening the two namespaces once drops `noop` to ~10ns and `simple_enum` to 2.2x. The published table in `benchmarks/README.md` predates this and overstates the scalar gap; the numbers below are all measured with the harness fixed, so they isolate the two changes above.

The Proxy existed to fall back between two BoltFFI packages: `run-bench.sh` packed the demo twice, once through the benchmark overlay and once plain, and `bench.mjs` imported both. They are the same crate — byte-identical wasm, the same 458 exported glue functions — so the fallback never resolved anything and the harness instantiated the module twice. The last commit imports the overlay package alone and drops the second pack, which also takes a full wasm build out of every benchmark run.

## Benchmarks

Node v26.5.0, x86_64 Linux, release. Both columns from a harness with the Proxy already removed:

| Benchmark | before | after | speedup | vs wasm-bindgen |
| --- | ---: | ---: | ---: | --- |
| find_direction | 771 ns | 10 ns | 77.1x | 154.2x slower → 1.67x |
| sum_user_scores_1k | 8.70 ms | 1.10 ms | 7.9x | 9.50x slower → 1.14x |
| count_active_users_100 | 818 us | 110 us | 7.4x | 8.00x slower → 1.15x |
| count_active_users_1k | 6.98 ms | 1.03 ms | 6.8x | 7.45x slower → 1.10x |
| sum_user_scores_100 | 738 us | 111 us | 6.6x | 7.34x slower → 1.15x |
| generate_directions_10k | 182 us | 68.7 us | 2.7x | 0.23x → 0.19x |

Absolute timings move a fair amount between runs on this machine, so the ratios against wasm-bindgen within a single run are the reliable figure. Benchmarks untouched by either change hold their ratio (`sum_i32_vec_10k` 1.00x, `generate_i32_vec_100k` 1.05x, `echo_bytes_64k` 1.17x).

## Testing

- `cargo test --workspace` — 46 groups, no failures. New tests cover the classification (`Option<enum>` → scalar, `repr(i64)` staying wire-encoded, `Option<record>` unchanged), the IR (wasm specializes, native does not, 64-bit repr does not), and the async return path, which has to pack the discriminant before it reaches the `f64` carrier — that test fails on the first revision of this branch.
- `examples/platforms/wasm` acceptance suite passes, which is what actually pins the values: `findDirection(2) === Direction.South`, `findDirection(99) === null`.
- Runtime suite is 104 tests, 67 of them new: the byte count is fuzzed against `TextEncoder` over the whole BMP, surrogate pairs, and 3000 random strings containing lone surrogates.

## Known follow-up

`ScalarEnumRepr::fits_in_f64` (boltffi_macros) and `Wasm32::scalar_option_enum` (boltffi_binding) must agree on which reprs are eligible; they do today and carry cross-references, but nothing enforces it automatically. If they ever drift, Rust would return an `f64` while TypeScript reads a buffer.

The remaining wasm gaps are `find_name` (5.8x) and `callback_1k` (4.4x). Both are the `takePackedBuffer` path: a fresh `ArrayBuffer` plus `DataView` per call, ~860ns of the 880ns that `Option<String>` costs. Pooling that reader is tempting but unsafe as things stand — `readI8Array`/`readU8Array` hand out views over the reader's buffer without copying, so a shared scratch would corrupt them silently. Left alone deliberately.


benchmarks/harnesses/wasm-bench/bench.mjs
```diff
@@ -6,14 +6,11 @@ const boltffiOverlay = await import('./build/generated/boltffi/node.js');
...
-const boltffi = new Proxy(boltffiOverlay, {
-  get(target, property, receiver) {
-    if (Reflect.has(target, property)) {
-      return Reflect.get(target, property, receiver);
-    }
-    return Reflect.get(boltffiDemo, property, receiver);
-  },
-});
+// Achatado uma vez, em vez de um Proxy consultado a cada chamada: um `get` trap
+// sobre um namespace de modulo custa ~200-300ns por acesso, o que dominava
```
engali94: can we translate it to english :)

boltffi_binding/src/ir/direction.rs
```diff
@@ -56,6 +56,14 @@ pub trait Direction:
     + serde::Serialize
     + for<'de> serde::Deserialize<'de>
 {
+    /// Whether a value in this direction is produced by Rust.
+    ///
+    /// `true` for [`OutOfRust`], where Rust returns the value; `false`
+    /// for [`IntoRust`], where foreign code produces it (a callback
+    /// return). Specializations that only one side implements gate on
+    /// this so the two sides cannot disagree on the ABI.
+    const PRODUCED_BY_RUST: bool;
```
engali94: i dont think this is necessary because the direction can be easly inferred from the Direction trait itself

benchmarks/harnesses/wasm-bench/bench.mjs
jlucaso1: sorry hehe. Fixing

engali94 (approved):
the speed up is impressive thanks @jlucaso1 !

jlucaso1:
@engali94 heads-up: I pushed 3a6fd8a4 after your approval, so it hasn't been reviewed.

It came out of the Codex feedback on #727. `run-bench.sh` packs the demo twice (overlay and plain) and `bench.mjs` imported both. They're the same crate: byte-identical wasm, same 458 exported glue functions. The `Proxy` fallback this PR removes never resolved anything, and the harness was instantiating the module twice for nothing.

The commit imports the overlay package alone and drops the second `pack wasm`. One less wasm build per run. Harness-only, benchmark still runs clean.

Happy to pull it into its own PR if you'd rather keep this one to what you approved.

Note: this makes `report-size.mjs` in #727 stale, since it still lists `boltffi-demo`. Whichever lands second needs a one-line follow-up.

PR #725, opened 2026-07-27
per-invocation source records: runtime and artifact reader

akesson:
First slice of #665. Inert plumbing only: nothing emits records yet, generated bindings are byte-identical.

- `boltffi_core`: const-buildable framing for per-invocation source records — one `BFFISRC1` frame per macro invocation in `__DATA,__boltffisrc` / `.boltffisrc`:

  ```
  magic | u64 LE length | package | version | module_path
        | u16 slot count | slot descriptors | u32-len JSON fragment
  ```

  plus the `TypeInfo`/`TypeDesc` identity traits capture slots project through (`<Point as TypeDesc<Tag>>::DESC` resolves a written reference at the referenced type's own definition site).
- `boltffi_binding`: the mirror frame reader and `aggregate_records`, which mints `$self::Name` ids from each record's `module_path!()`, resolves `$slot:N` leaves against the defining fragments, collapses identical duplicates (multi-artifact reads stay safe), and refuses incomplete sets so consumers fall back to the legacy path.
- `boltffi_bindgen`: `read_source` — reads records from every artifact of a plain cargo build, dependency rlibs and wasm32 objects included (`object` gains its `wasm` read feature; without it wasm members in rlibs were silently skipped).

Part of #665. Next: emission on every macro entry point, then the discovery flip.

### DIFF
--- a/boltffi_binding/src/ir/source_fragment.rs
+++ b/boltffi_binding/src/ir/source_fragment.rs
@@ -102,14 +102,14 @@ pub fn aggregate_records(
     }
 
     let mut kinds = HashMap::new();
-    let mut unique = HashMap::new();
-    for (fragment, _) in &fragments {
+    let mut unique: HashMap<String, (&SourceFragment, &[TypeNode])> = HashMap::new();
+    for (fragment, slots) in &fragments {
         let id = fragment_id(fragment).to_owned();
         match unique.entry(id.clone()) {
             Entry::Vacant(entry) => {
-                entry.insert(fragment.clone());
+                entry.insert((fragment, slots));
             }
-            Entry::Occupied(entry) if entry.get() == fragment => {}
+            Entry::Occupied(entry) if *entry.get() == (fragment, slots.as_slice()) => {}
             Entry::Occupied(_) => {
                 return Err(SourceFragmentError::DuplicateDeclaration { id });
             }
@@ -763,6 +763,19 @@ mod tests {
             .expect_err("conflicting duplicates fail");
         assert!(
             matches!(error, SourceFragmentError::DuplicateDeclaration { id } if id == "demo::geometry::Point")
+        );
+    }
+
+    #[test]
+    fn rejects_duplicate_declarations_whose_slots_differ() {
+        let json = record_fragment(vec![FieldDef::new(name("start"), slot_leaf(0, "Alias"))]);
+        let first = raw("demo", &[r#"{"prim":"f32"}"#], json.clone());
+        let second = raw("demo", &[r#"{"prim":"f64"}"#], json);
+
+        let error = aggregate_records(&[first, second], PackageInfo::new("demo", None))
+            .expect_err("identical fragments with differing slots fail");
+        assert!(
+            matches!(error, SourceFragmentError::DuplicateDeclaration { id } if id == "demo::Route")
         );
     }
 
--- a/boltffi_binding/src/ir/source_fragment.rs
+++ b/boltffi_binding/src/ir/source_fragment.rs
@@ -1,0 +1,778 @@
+//! Interpreting per-invocation source records into one aggregated [`SourceContract`].
+//!
+//! A record's JSON payload is one [`SourceFragment`]: the `boltffi_ast` declaration the
+//! macro invocation saw, with two kinds of placeholder the macro cannot fill in itself.
+//! Ids are written `$self::Name` and minted here against the record's `module_path!()`. Named type references are provisional [`TypeExpr::Record`] leaves
+//! whose id is `$slot:N`, replaced here by the type node the referenced type's own
+//! definition site resolved through rustc.
+
+use std::collections::HashMap;
+use std::collections::hash_map::Entry;
+
+use boltffi_ast::{
+    ClassDef, ClassId, ConstantDef, ConstantId, CustomTypeDef, EnumDef, EnumId, FieldDef,
+    FunctionDef, FunctionId, MethodDef, NamePart, PackageInfo, Path, PathSegment, Primitive,
+    RecordDef, RecordId, ReturnDef, SourceContract, StreamDef, StreamId, TraitDef, TraitId,
+    TypeExpr, VariantPayload,
+};
+use serde::{Deserialize, Serialize};
+
+use crate::ir::source_record::RawSourceRecord;
+
+/// Id prefix marking a declaration's own identity, minted from the record's module path.
+pub const SELF_ID: &str = "$self";
+
+/// Id prefix marking a slot-deferred type reference inside a fragment.
+pub const SLOT_ID_PREFIX: &str = "$slot:";
+
+/// One declaration captured by one macro invocation.
+#[derive(Clone, Debug, Eq, Hash, PartialEq, Serialize, Deserialize)]
+#[serde(tag = "kind", rename_all = "snake_case")]
+pub enum SourceFragment {
+    /// A `#[data]` struct.
+    Record(RecordDef),
+    /// A `#[data]` enum.
+    Enum(EnumDef),
+    /// An `#[export]` free function.
+    Function(FunctionDef),
+    /// An `#[export]`ed class-style object.
+    Class(ClassDef),
+    /// An `#[export]`ed callback trait.
+    Trait(TraitDef),
+    /// A stream export.
+    Stream(StreamDef),
+    /// An `#[export]`ed constant.
+    Constant(ConstantDef),
+    /// A `custom_type!` registration.
+    Custom(CustomTypeDef),
+}
+
+/// One resolved type node from a slot descriptor.
+#[derive(Clone, Debug, Eq, Hash, PartialEq, Serialize, Deserialize)]
+#[serde(untagged)]
+pub enum TypeNode {
+    /// A named declaration, by canonical id.
+    Id {
+        /// Canonical id of the referenced declaration.
+        id: String,
+    },
+    /// A primitive or builtin leaf.
+    Prim {
+        /// Rust spelling of the leaf type.
+        prim: String,
+    },
+    /// A container shape with type arguments.
+    Shape {
+        /// Shape name, such as `Vec` or `HashMap`.
+        shape: String,
+        /// Type arguments in declaration order.
+        args: Vec<TypeNode>,
+    },
+}
+
+/// Aggregates decoded source records into one source contract.
+///
+/// The contract's declarations are sorted by id within each family, so the result does
+/// not depend on link order.
+pub fn aggregate_records(
+    records: &[RawSourceRecord],
+    package: PackageInfo,
+) -> Result<SourceContract, SourceFragmentError> {
+    let mut fragments = Vec::new();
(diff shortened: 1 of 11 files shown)


boltffi_binding/src/ir/source_fragment.rs
```diff
@@ -0,0 +1,778 @@
...
+            {
+                *owner = ClassId::new(value);
+            }
+        }
+        SourceFragment::Constant(def) => {
+            if let Some(value) = minted(def.id.as_str(), module) {
+                def.id = ConstantId::new(value);
+            }
+        }
+        SourceFragment::Custom(_) => {}
```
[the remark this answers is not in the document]
akesson: Slice boundary: nothing emits custom fragments yet. Their id minting (plus converter paths) lands with the emission PR.

boltffi_binding/src/ir/source_fragment.rs
```diff
@@ -0,0 +1,778 @@
...
+                && let Some(value) = minted(owner.as_str(), module)
+            {
+                *owner = ClassId::new(value);
+            }
+        }
+        SourceFragment::Constant(def) => {
+            if let Some(value) = minted(def.id.as_str(), module) {
+                def.id = ConstantId::new(value);
+            }
+        }
```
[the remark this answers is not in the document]
akesson: Owner minting rides the emission PR: class owners are minted there, and record/enum owners are assigned when their method-block fragment merges into the resolved target. A standalone constant fragment with a record owner is never emitted.

boltffi_binding/src/ir/source_fragment.rs
```diff
@@ -0,0 +1,778 @@
...
+
+    let mut kinds = HashMap::new();
+    let mut unique = HashMap::new();
+    for (fragment, _) in &fragments {
+        let id = fragment_id(fragment).to_owned();
+        match unique.entry(id.clone()) {
+            Entry::Vacant(entry) => {
+                entry.insert(fragment.clone());
+            }
+            Entry::Occupied(entry) if entry.get() == fragment => {}
```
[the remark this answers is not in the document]
akesson: Fixed in 49e1154e — slot descriptors join the identity comparison, so this now fails loudly as a duplicate.

boltffi_binding/src/ir/source_fragment.rs
```diff
@@ -0,0 +1,778 @@
...
+            Entry::Occupied(_) => {
+                return Err(SourceFragmentError::DuplicateDeclaration { id });
```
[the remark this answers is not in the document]
akesson: Half right. Data-type impls ride method-block fragments whose ids include member names, so complementary blocks never collide — they merge into the resolved target. Split class impls are a real gap though: the emission PR will merge class fragments the way the scanner merges impl blocks.

boltffi_binding/src/ir/source_fragment.rs
```diff
@@ -0,0 +1,778 @@
...
+fn declared_kind(fragment: &SourceFragment) -> Option<DeclaredKind> {
+    match fragment {
+        SourceFragment::Record(_) => Some(DeclaredKind::Record),
+        SourceFragment::Enum(_) => Some(DeclaredKind::Enum),
+        SourceFragment::Class(_) => Some(DeclaredKind::Class),
+        SourceFragment::Custom(_) => Some(DeclaredKind::Custom),
+        SourceFragment::Function(_)
+        | SourceFragment::Trait(_)
+        | SourceFragment::Stream(_)
+        | SourceFragment::Constant(_) => None,
```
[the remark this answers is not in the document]
akesson: Slice boundary: traits join the kind map in the emission PR, alongside the callback capture that emits them.

boltffi_core/src/capture/type_info.rs
```diff
@@ -0,0 +1,244 @@
...
+    };
+}
+
+primitive_desc!(
+    bool, u8, u16, u32, u64, i8, i16, i32, i64, usize, isize, f32, f64
+);
+
+impl<Tag> TypeDesc<Tag> for String {
+    const DESC: DescBuf = DescBuf::primitive("String");
+}
```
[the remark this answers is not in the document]
akesson: Builtin spellings (Duration, std::time::Duration, ...) are recognized at capture before any slot is minted, same as the scanner — a Duration slot can only arise from an alias spelling, which the scanner rejects too. Going beyond scanner parity is deliberately out of scope for this series.

boltffi_core/src/capture/type_info.rs
```diff
@@ -0,0 +1,244 @@
...
+            "two-argument shapes separate their arguments"
+        );
+    }
+
+    #[test]
+    fn descriptor_strings_promote_to_static_slots() {
+        const DESC: &DescBuf = &<Points as TypeDesc<ProbeTag>>::DESC;
+        const SLOT: &str = DESC.as_str();
+        const SLOTS: &[&str] = &[SLOT];
+        const LEN: usize = crate::capture::record_len("p", "", "m", SLOTS, b"{}");
```
engali94: crate::capture:: 
path being repetitive here can we define it in the top of the mod

boltffi_core/src/capture/type_info.rs
```diff
@@ -0,0 +1,244 @@
...
+    label = "no canonical id",
+    note = "annotate `{Self}` with `#[data]`, or declare it with `custom_type!` if it is foreign"
+)]
+pub trait TypeDesc<Tag> {
+    /// One JSON type node: `{"id":..}`, `{"prim":..}`, or `{"shape":..,"args":[..]}`.
+    const DESC: DescBuf;
+}
+
+/// Capacity of a [`DescBuf`]; deep nesting past this is a compile error.
+pub const DESC_CAPACITY: usize = 1024;
```
engali94, line 30: question: wondering how deep nesting this can hold?

boltffi_core/src/capture/type_info.rs
```diff
@@ -0,0 +1,244 @@
...
+            .push(b"\"}")
+    }
+
+    /// Builds the node for a one-argument container shape.
+    pub const fn shape1(name: &str, arg: Self) -> Self {
+        Self::shape_open(name).concat(arg).push(b"]}")
+    }
+
+    /// Builds the node for a two-argument container shape.
+    pub const fn shape2(name: &str, first: Self, second: Self) -> Self {
```
engali94: it took a while to click for me
i guess if we rename it to `shape_with_two_args` would be easier to understand  and similarly for `shape1`

boltffi_core/src/capture/type_info.rs
```diff
@@ -0,0 +1,244 @@
...
+primitive_desc!(
+    bool, u8, u16, u32, u64, i8, i16, i32, i64, usize, isize, f32, f64
+);
+
+impl<Tag> TypeDesc<Tag> for String {
+    const DESC: DescBuf = DescBuf::primitive("String");
+}
+
+impl<Tag, T: TypeDesc<Tag>> TypeDesc<Tag> for Vec<T> {
+    const DESC: DescBuf = DescBuf::shape1("Vec", T::DESC);
```
engali94: thats clever!

boltffi_core/src/capture/type_info.rs
```diff
@@ -0,0 +1,244 @@
...
+}
+
+/// Structural descriptor of a type at an FFI use site, composable in const context so
+/// aliases to containers resolve through the compiler.
+#[diagnostic::on_unimplemented(
+    message = "`{Self}` has no canonical id, so it cannot cross the FFI boundary",
+    label = "no canonical id",
+    note = "annotate `{Self}` with `#[data]`, or declare it with `custom_type!` if it is foreign"
+)]
+pub trait TypeDesc<Tag> {
```
engali94, line 24: thats clever! it will help us easily export custom types than what we do today!

engali94 (approved):
very well done!
one thing i wanted to highlight currently BoltFFI does this 
```rust
#[data]
struct Point {
    x: f64,
}

#[data(impl)] 
impl Point {
    pub const ZERO: Self = Self { x: 0.0 };

    pub fn length(&self) -> f64 {
        self.x.abs()
    }
}
```

Those are two macro invocations describing the same Point. so the new the `aggregate_records` needs somehow maybe to co-relate them

boltffi_core/src/capture/type_info.rs
akesson: Done in 5ec4100e.

boltffi_core/src/capture/type_info.rs
akesson: Each container level costs ~25 bytes (`{"shape":"Vec","args":[` ... `]}`), so 1024 holds roughly 40 nested levels around a leaf — real signatures run 2-4 deep. Overflow is a compile error ("type descriptor exceeds the capture buffer"), never silent truncation, and the capacity is a one-line bump if it ever bites.

boltffi_core/src/capture/type_info.rs
akesson: Renamed in 5ec4100e.

boltffi_core/src/capture/type_info.rs
```diff
@@ -64,12 +64,12 @@ impl DescBuf {
...
 
     /// Builds the node for a one-argument container shape.
-    pub const fn shape1(name: &str, arg: Self) -> Self {
+    pub const fn shape_with_one_arg(name: &str, arg: Self) -> Self {
         Self::shape_open(name).concat(arg).push(b"]}")
     }
 
     /// Builds the node for a two-argument container shape.
-    pub const fn shape2(name: &str, first: Self, second: Self) -> Self {
+    pub const fn shape_with_two_args(name: &str, first: Self, second: Self) -> Self {
```
engali94, line 72: it took a while to click for me
i guess if we rename it to `shape_with_two_args` would be easier to understand  and similarly for `shape1`

engali94: stale comment. github is buggy

akesson:
Yes, designed for: `#[data(impl)]` emits a *methods* fragment rather than a second `Point` declaration, and its target is a compiler-resolved slot (`<Point as TypeDesc<_>>::DESC`), so aggregation merges the block into the `#[data]`-owned declaration by canonical id — no name guessing. Lands with the emission PR.

engali94 (approved):

akesson:
@engali94 I'll go ahead and merge this since you approved it. Is that ok? Is it something I can always do on approved PRs?

boltffi_binding/src/ir/source_fragment.rs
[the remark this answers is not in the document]
akesson: Correction after testing: split class impls have never compiled — the legacy expansion emits `boltffi_{class}_free` per impl block, so two `#[export] impl` blocks for one class are E0428 on main today (probe crate confirms). The records path cannot regress a pattern rustc rejects, so aggregation keeps the loud duplicate error; merging belongs with the class-identity-site question if that pattern is ever enabled.

PR #728, opened 2026-07-27
stop inlining the error path into every generated wasm shim

jlucaso1:
The generated export shims each carry their own inlined copy of the error path. Removing that takes the demo's wasm from 885,896 to 788,479 bytes (-11.0%), with no API or behaviour change.

## What was happening

`set_last_error` (`boltffi_core/src/status.rs`) is a small non-`#[cold]` generic, so LLVM inlines the whole body — thread-local access, the `RefCell` borrow-flag check, a `String` allocation, and the unwind-free panic trailer — into every function that can fail. On `examples/demo` that is ~300 exported shims.

Counting call targets across the shims in the baseline module:

| inlined callee | shims carrying a copy |
| --- | ---: |
| `core::cell::panic_already_borrowed` | 308 |
| `__rust_dealloc` | 350 |
| `__rust_no_alloc_shim_is_unstable_v2` | 332 |
| `alloc::raw_vec::handle_error` | 309 |
| `__rust_alloc` | 239 |
| `alloc::fmt::format_inner` | 188 |

Generated shims account for 267 KiB of the 536 KiB code section, averaging 463 bytes each — most of which is failure handling that never runs.

## Three changes

Each measured on `pack wasm --release` of `examples/demo`:

**1. `set_last_error` is `#[cold] #[inline(never)]`.** 885,896 → 814,448 bytes. `panic_already_borrowed` callers drop from 308 to 15. Error paths are cold by definition, and the happy path benefits from the lower register pressure.

**2. `concat!` where the message is a compile-time constant.** 26 sites whose only interpolation was `stringify!(#ident)` now reference a `&'static str` instead of building a `String`. → 810,435 bytes.

**3. Outline the failures that carry a runtime value.** Three `#[cold]` helpers — `set_last_error_len`, `set_last_error_display`, `set_last_error_debug` — replace 36 inlined `format!` sites. Taking `&dyn Display` / `&dyn Debug` also collapses what was one monomorphisation per error type into a single instantiation. → 788,479 bytes.

| | bytes |
| --- | ---: |
| before | 885,896 |
| after | **788,479** |

## Messages are unchanged

What changes is how many copies ship, not what they say. Counted in the wasm data section, before → after:

| message | before | after |
| --- | ---: | ---: |
| `wire decode failed` | 77 | 1 |
| `null pointer with non-zero length` | 86 | 1 |
| `custom conversion failed` | 6 | 1 |
| `" (buf_len="` | 174 | 6 |

Two message families (`invalid UTF-8`, `invalid optional scalar payload`) appear zero times both before and after — the demo does not reach those paths, so nothing was lost there either.

## Testing

- `cargo test --workspace` — 46 groups, no failures. The token-stream expectations in `boltffi_macros` were updated to match the new expansion; they assert on the same messages.
- `examples/platforms/wasm` acceptance suite passes.

## Not included

Shortening the exported symbol names would recover up to another 54 KiB, but the name section is stripped from release artifacts, so export names are the only human-readable symbols left in a shipped `.wasm`. Making them opaque would darken browser stack frames and post-hoc inspection of production artifacts, so it is left out.

Two smaller things found while measuring, worth their own issues rather than bundling here: ~94 async-completion shims carry an identical `assert_failed::<bool,bool>` trailer (~2.4 KiB total), and the build embeds absolute paths from the machine that produced it (`/home/<user>/.cargo/registry/...`, ~2.6 KiB) — more a privacy matter than a size one.

### DIFF
--- a/boltffi/src/lib.rs
+++ b/boltffi/src/lib.rs
@@ -41,7 +41,8 @@ pub mod __private {
         FfiSpan, FfiStatus, InternedString, InternedStringPool, InternedStringRepr,
         NativeCallbackOwner, Passable, RustFutureContinuationCallback, RustFutureHandle,
         StreamContinuationCallback, StreamPollResult, SubscriptionHandle, VecTransport, WaitResult,
-        WirePassable, rustfuture, set_last_error, take_last_error, wire,
+        WirePassable, rustfuture, set_last_error, set_last_error_debug, set_last_error_display,
+        set_last_error_len, take_last_error, wire,
     };
     #[cfg(target_arch = "wasm32")]
     pub use boltffi_core::{
--- a/boltffi_core/src/lib.rs
+++ b/boltffi_core/src/lib.rs
@@ -66,7 +66,10 @@ pub use runtime::subscription::{
     SubscriptionHandle, WaitResult,
 };
 pub use safety::catch_ffi_panic;
-pub use status::{FfiStatus, clear_last_error, set_last_error, take_last_error};
+pub use status::{
+    FfiStatus, clear_last_error, set_last_error, set_last_error_debug, set_last_error_display,
+    set_last_error_len, take_last_error,
+};
 
 pub use types::{FfiBuf, FfiError, FfiOption, FfiSlice, FfiSpan, FfiString};
 pub use wasm::WASM_ABI_VERSION;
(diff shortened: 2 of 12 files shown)


engali94 (approved):
nice cleanup!

PR #735, opened 2026-07-28
decode encoded returns in place instead of copying them out

jlucaso1:
Encoded returns copy their payload out of wasm memory before the codec reads it. Decoding in place instead takes `find_name` from ~956ns to ~354ns.

## What was happening

`takePackedBuffer` allocates a fresh `ArrayBuffer` and a new `DataView` on every call:

```ts
const bytes = new Uint8Array(this._memory.buffer, pointer, length);
const copy = bytes.slice();
this.freePacked(pointer, length);
return new WireReader(copy.buffer, 0);
```

For a small record that allocation is most of the cost — the work happens before the codec reads a single field. Every `Option<T>`, record, and `Vec<record>` return goes through it.

## The change

`readPackedBuffer(packed, read)` lends the reader wasm memory and frees the payload once `read` returns, so no copy. The generator emits it in place of the two-statement form:

```ts
// before
const __boltffiReader = _module.takePackedBuffer(_exports.…(ptr, len));
return __boltffiReader.readOptional(() => NodeSummaryCodec.decode(__boltffiReader));

// after
return _module.readPackedBuffer(_exports.…(ptr, len),
  (__boltffiReader) => __boltffiReader.readOptional(() => NodeSummaryCodec.decode(__boltffiReader)));
```

### Why this is safe now, having been unsafe in #728

#728 flagged this path and deliberately left it alone:

> Pooling that reader is tempting but unsafe as things stand — `readI8Array`/`readU8Array` hand out views over the reader's buffer without copying, so a shared scratch would corrupt them silently.

That objection was correct and is what this addresses. `WireReader` now tracks whether it borrows wasm memory, and the three reads that would hand out a view over it — `readI8Array`, `readU8Array`, `readTypedArray` — copy in that mode. Everything else already produced owned values: scalars are copied by construction, `readString` builds a new string, `readBytes` already called `.slice()`.

Three further things the copy used to give for free, restored explicitly in the second commit:

- **Bounds.** The copy was exactly `length` bytes, so an over-long nested length threw. A borrowed reader's view now spans exactly the payload, and the reads that build a typed array — which bypass the view's own check — go through an explicit one.
- **Escape.** The reader is detached after `read` returns, so one that outlives the callback throws instead of reading freed memory.
- **Reentrancy.** A nested `readPackedBuffer` gets its own reader rather than resetting the outer one's. Generated codecs never nest, but the method is public.

A `read` callback returning a promise is still the caller's problem: the payload is freed when `read` returns, not when the promise settles. Generated codecs are synchronous.

## Numbers

`examples/demo`, median of 5 repetitions per case, two full paired rounds (rebuild `main`, measure, rebuild branch, measure):

| case | before | after | |
| --- | ---: | ---: | --- |
| `find_name` (`Option<String>`) | 956 ns | **354 ns** | 2.70x |
| `generate_particles_10k` (`Vec<record>`) | 976 us | **759 us** | 1.29x |
| `sum_trade_volumes_10k` | 3.26 ms | **2.70 ms** | 1.21x |
| `echo_string_small` (control, different path) | 262 ns | 248 ns | unchanged |

The gain grows with payload size, because the avoided copy grows with it. `find_name` is the case #728 used as its example of this cost.

### On reproducing these

**Please run them yourself.** `benchmarks/harnesses/wasm-bench` has enough run-to-run variance that a single run of each side is not trustworthy: at one point, comparing one baseline run against one patched run made `generate_particles_10k` look **2.86x worse**, when repeated paired measurement shows it better. That case alone ranged 596us–1308us within a single baseline round.

The table therefore comes from a focused probe taking the median of 5 repetitions, run twice with a full rebuild of each side in between, with `echo_string_small` carried as a control that this change should not touch. I would not trust — and am not claiming — a global "everything got faster" figure from this harness. What I am claiming is the cases above, which held across both rounds, and the mechanism, which is straightforward: one fewer allocation and copy per encoded return.

An earlier revision of this branch measured `find_name` lower still, before the payload bound was added back. Bounding the view costs a `DataView` per call that the unbounded version could reuse; that trade seemed clearly worth making.

## Testing

- `cargo test --workspace` — no failures. Two snapshot expectations in `boltffi_backend` were updated: the generated glue changed shape, not semantics.
- Runtime suite, 115 tests, including a new `borrowed.test.ts`: it writes a payload, decodes it through a borrowed reader, then overwrites the source bytes, so anything still aliasing them fails. Covers every read that could alias, plus the bounds and detach behaviour.
- `examples/platforms/wasm` acceptance suite passes, which is what pins the decoded values.

The borrowed-mode copies are the part most worth reviewing: if a read that returns a view over the reader's buffer were missed, it would produce a use-after-free. I believe `readI8Array`, `readU8Array` and `readTypedArray` are the complete set — `readBoolArray` builds a new array via `Array.from`, and `readBytes` already copies — but a second pair of eyes on that list is the highest-value review here.

### DIFF
--- a/boltffi_backend/src/target/typescript/mod.rs
+++ b/boltffi_backend/src/target/typescript/mod.rs
@@ -1135,12 +1135,12 @@ mod tests {
             "const __boltffi_value_writer = _module.allocWriter(PointCodec.size(value));"
         ));
         assert!(browser.contents().contains(
-            "const __boltffiReader = _module.takePackedBuffer((_exports.boltffi_function_demo_echo_point"
+            "return _module.readPackedBuffer((_exports.boltffi_function_demo_echo_point"
         ));
         assert!(
             browser
                 .contents()
-                .contains("return PointCodec.decode(__boltffiReader);")
+                .contains("(__boltffiReader) => PointCodec.decode(__boltffiReader)")
         );
         assert!(
             browser
--- a/boltffi_backend/src/target/typescript/render/function.rs
+++ b/boltffi_backend/src/target/typescript/render/function.rs
@@ -1556,19 +1556,21 @@ impl Return {
                     .into_iter()
                     .collect::<ArgumentList>(),
             ))],
-            ReturnConversion::Encoded { reader, decode } => vec![
-                Statement::constant(
-                    reader.clone(),
-                    Expression::call(
-                        Expression::identifier(Identifier::known("_module")),
-                        Identifier::known("takePackedBuffer"),
-                        [call.cast(TypeName::bigint())]
-                            .into_iter()
-                            .collect::<ArgumentList>(),
-                    ),
-                ),
-                Statement::return_value(decode.clone()),
-            ],
+            // Decoded in place rather than copied out: `readPackedBuffer` lends
+            // the reader wasm memory and frees the payload once the codec
+            // returns, which avoids an ArrayBuffer and a DataView per call.
+            ReturnConversion::Encoded { reader, decode } => {
+                vec![Statement::return_value(Expression::call(
+                    Expression::identifier(Identifier::known("_module")),
+                    Identifier::known("readPackedBuffer"),
+                    [
+                        call.cast(TypeName::bigint()),
+                        Expression::parameter_lambda(reader.clone(), decode.clone()),
+                    ]
+                    .into_iter()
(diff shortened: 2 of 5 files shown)


engali94 (approved):
NICE!

PR #740, opened 2026-07-28
perf(wasm): stop paying per-call costs at the wasm boundary

jlucaso1:
Three things the runtime did on every call across the boundary, none of which it needs to do. They are in one PR because their costs interlock: removing the accessor from `reset` buys nothing while the reader still allocates a `DataView` per call, so measured apart one number understates and the other overstates.

## 1. The memory caches revalidated through an accessor

`getBytes()`, `getView()`, `getI32()` and the rest checked `cached.buffer !== this._memory.buffer`. Both sides are accessors, not field loads.

| test | over floor |
| --- | ---: |
| `cached.buffer !== memory.buffer` | 15.91 ns |
| ↳ `memory.buffer` alone | 8.84 ns |
| ↳ `cached.buffer` alone | 6.48 ns |
| `cached.byteLength === 0` | **0.52 ns** |

This is paid by every runtime operation — `allocWireBytes` calls two of these getters, and the pair cost more than its two wasm calls put together (31 ns against 23 ns). A CPU profile shows `get buffer ()` as its own entry at 3.5% of the echo path, which is the evidence that does not depend on trusting a microbenchmark.

A detached typed array reports `byteLength === 0`, so the cache can test itself. Two things make this less obvious than it looks, both commented in the diff:

- **`DataView.prototype.byteLength` throws when detached** rather than reporting 0, so that cache cannot self-test and carries a `Uint8Array` probe alongside it.
- **Growing a *shared* `WebAssembly.Memory` does not detach the old buffer.** The old identity comparison caught that; this one would not. Nothing in the repo creates shared memory today, so it is dormant, but it is a trap for whoever adds threads and I would rather flag it than have it found later.

A module with zero initial pages reports `byteLength === 0` forever and remaps on every call — correct, just slow.

## 2. Unpacking the packed i64 allocated BigInts

`Number(packed & 0xffff_ffffn)` allocates an intermediate BigInt, and the shift allocates another. Every packed return path did this.

| approach | cost |
| --- | ---: |
| mask + shift | 39.3 ns |
| `BigInt.asUintN` | 36.1 ns |
| aliased `BigUint64Array`/`Uint32Array` | **12.1 ns** |

The ten other returns that split the value by hand now share `unpackPacked` instead of repeating it.

Note that aliasing the two views is **host-endian**, where the BigInt arithmetic it replaces was not. Which half is which is settled once at load, into `PACKED_LOW`/`PACKED_HIGH`. A reviewer who sees the aliasing and not the constants would be right to call it a bug.

## 3. The borrowed reader allocated a `DataView` per call

#735 bounded the reader with `new DataView(buffer, ptr, len)`. Correct, but an allocation per call, and it defeats the reuse `reset()` was there to provide.

| approach | cost |
| --- | ---: |
| `new DataView(buf, ptr, len)` per call | 49.4 ns |
| cached view + `limit` field | **3.6 ns** |

The guarantee is not automatically the same, and getting it wrong is the risk in this PR. A payload-relative bounded view bounds *both* ends; a `limit` only bounds the top. `skip()` now goes through `take()` as well, so a negative argument can no longer walk below the payload — it could before this change, and there is a test for it.

The trade-off, stated plainly: the bounded view got its per-read check free from the view, while `limit` pays an explicit compare on every read — **+0.306 ns per read**. Break-even is somewhere around a hundred-odd reads per call. Records with a handful of fields win clearly; a single call decoding a very large array pays a little.

## Numbers

Allocation, measured per path against `main`. These are counts rather than timings, and they reproduce:

| path | bytes/call | scavenges per 200k calls |
| --- | --- | --- |
| decode a record | 575.0 → **381.3** (−34%) | 55 → **37** |
| echo bytes | 480.1 → **337.4** (−30%) | 46 → **33** |
| encode a record | 480.7 → **337.4** (−30%) | 46 → **33** |

End-to-end CPU, as a multiple of a raw wasm call measured in the same process: a record decode went from **2.54x to 1.94x**, so the JS glue above the wasm call fell from 1.54 to 0.94 — roughly **−39% of the glue**.

### Please run these yourself

My machine was heavily loaded throughout (loadavg 12–33). The allocation table above is trustworthy — it counts objects. The CPU figure is the only end-to-end timing here that exceeded the noise floor; sibling PRs measured on the same machine showed ±8% swings on code paths they do not touch, which is why this PR claims one CPU number and not a table of them.

Measurement method, if it is useful: variants interleaved in a single process, each run carrying a raw wasm call as a control so results are multiples of that control rather than wall-clock, and minimum over many repetitions.

## Testing

- Runtime suite, 126 tests. New `reader-bounds.test.ts` covers both ends of the payload (scalar, array, string, typed array, and `skip` in both directions), that borrowed reads copy rather than alias, that a reader outliving its payload throws, and that reads stay correct after wasm memory grows.
- New `test/support/fake-module.ts` builds a `BoltFFIModule` over a plain `WebAssembly.Memory` and a bump allocator, so the packed-return paths can be tested without a wasm build. There was no fixture for this before.
- I checked the growth test fails without the fix: with the detach check disabled it fails exactly at the growth point.
- `examples/platforms/wasm` acceptance suite passes, which is what pins the decoded values end to end.
- No Rust changes in this PR, so the cargo suites are unaffected.

Touches `wire.ts` in the same region as #739 (short ascii strings); whichever lands second needs a trivial rebase.

### DIFF
--- a/runtime/typescript/src/module.ts
+++ b/runtime/typescript/src/module.ts
@@ -4,6 +4,10 @@ import { StreamPollManager } from "./stream.js";
 
 const EMPTY_BUFFER = new ArrayBuffer(0);
 
+const LITTLE_ENDIAN = new Uint8Array(new Uint32Array([1]).buffer)[0] === 1;
+const PACKED_LOW = LITTLE_ENDIAN ? 0 : 1;
+const PACKED_HIGH = LITTLE_ENDIAN ? 1 : 0;
+
 const FFI_BUF_DESCRIPTOR_SIZE = 16;
 const FFI_STATUS_SIZE = 4;
 const OPTION_F64_NONE = 0xffff_ffff_ffff_ffffn;
@@ -197,6 +201,11 @@ export class BoltFFIModule {
   private _cachedF32: Float32Array | null = null;
   private _cachedF64: Float64Array | null = null;
   private _cachedView: DataView | null = null;
+  private _viewProbe: Uint8Array | null = null;
+  private _memoryBuffer: ArrayBuffer | null = null;
+  private _packedStorage = new ArrayBuffer(8);
+  private _packedBits = new BigUint64Array(this._packedStorage);
+  private _packedHalves = new Uint32Array(this._packedStorage);
   private _optionF64Storage = new ArrayBuffer(8);
   private _optionF64Bits = new BigUint64Array(this._optionF64Storage);
   private _optionF64Values = new Float64Array(this._optionF64Storage);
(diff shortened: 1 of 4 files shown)


engali94 (approved):
wow thats very nice
thank you @jlucaso1 !

PR #764, opened 2026-07-30
Safely convert unexpected callback errors

ahmedre:


ahmedre:
before

```swift
  do {
      return try wrapper.impl.replan(...)
  } catch {
      let bytes = boltffiEncode { writer in
          writer.writeI32(
              (error as? ReplanningError)!.rawValue
          )
      }
      return boltffi_buf_from_bytes(bytes)
  }
```

after

```swift
  do {
      return try wrapper.impl.replan(...)
  } catch let error as ReplanningError {
      // Normal declared error: keep the existing wire format.
      let bytes = boltffiEncode { writer in
          writer.writeI32(error.rawValue)
      }
      return boltffi_buf_from_bytes(bytes)
  } catch {
      // Different Swift error: send it back without casting.
      let bytes = boltffiEncodeUnexpectedCallbackError(error)
      return boltffi_buf_from_bytes(bytes)
  }
```

engali94 (approved):

PR #765, opened 2026-07-30
perf(wasm): validate the writer's memory cache without the accessor

jlucaso1:
`WireWriter.currentView()` called `allocator.buffer()` on every write. That callback is `() => this._memory.buffer`, and `WebAssembly.Memory.prototype.buffer` is an **accessor**, not a field — the same cost #740 took off the read path, still on the write path.

## How many times

Counted by intercepting the getter on `WebAssembly.Memory.prototype`, so this is a count rather than a timing and does not depend on machine load. The interceptor was validated against a known number of accesses first.

| case | `memory.buffer` reads per call |
| --- | ---: |
| `countActiveUsers(1k)` | **21 001** |
| `sumParticleMasses(1k)` | 10 001 |
| `processLocations(1k)` | 7 000 |
| `sumUserScores(100)` | 2 101 |
| `noop`, `findDirection`, `sumI32Vec(10k)` | 0 |

One read per scalar field written. The cases that read zero are the ones with no wire-encoded parameter — which is why this went unnoticed: a benchmark without a record parameter never reaches the code.

## The change

The cached view and buffer are validated by a detachment probe instead of by calling `allocator.buffer()` and comparing. Growing wasm memory detaches the old `ArrayBuffer`, so a cached `Uint8Array` over it reports `byteLength === 0`, and that is a plain field read.

The probe has to be a `Uint8Array`. `DataView.prototype.byteLength` **throws** when its buffer is detached rather than reporting 0, so the cached view cannot probe itself — the same trap #740 hit.

## Numbers

`examples/demo` through `benchmarks/harnesses/wasm-bench`, three runs per side, medians. boltffi nanoseconds, and the ratio to wasm-bindgen measured in the same process:

| case | before | after | vs wasm-bindgen |
| --- | ---: | ---: | --- |
| `process_locations_1k` | 134 us | **58 us** | 1.46x → **3.85x** faster |
| `sum_particle_masses_1k` | 211 us | **114 us** | 1.12x → **2.01x** faster |
| `noop` (control) | 6 ns | 6 ns | unchanged |

Neither case's three runs overlap between the two sides. `noop` is carried as a control that this change cannot touch, and it does not move.

Two more cases improve but I am not claiming a figure for them: `sum_user_scores_100` reads 1.06–1.10x faster after, against a baseline whose own three runs disagreed on the sign (1.30x slower, 1.12x slower, 1.14x faster), and `count_active_users_1k` improves in direction while neither side settles — one baseline run came in at 2.95 ms against 1.01 ms for another.

A caveat on conditions: the machine was at loadavg 6–10 throughout, so the absolutes above are inflated and only the ratios and the before/after comparison are trustworthy. The read counts are unaffected by that.

## Testing

- Runtime suite 160 tests, `tsc --noEmit` clean. Four new tests in `test/writer-growth.test.ts` grow wasm memory while a writer is live: between two writes, in the middle of encoding a record, immediately before `writeBytes`, and on a region-bound writer sharing one allocator with another.
- **All four fail with the probe check replaced by an unconditional cache hit**, with `TypeError: Cannot perform DataView.prototype.setUint32 on a detached or out-of-bounds ArrayBuffer`. I checked that by making the change and running them, not by inspection.
- `examples/platforms/wasm` acceptance suite passes.
- `cargo test --workspace`, `cargo clippy --workspace --all-targets`, `cargo fmt --all` clean.

## Shared memory — raised in review, now fixed

The first version of this branch left shared memory as a documented caveat. That was too weak, and the review was right to push: growing a shared `WebAssembly.Memory` replaces `memory.buffer` without detaching the old `SharedArrayBuffer`, so the probe stays nonzero and the stale view is reused — and if `realloc` grows the memory and moves the allocation past the old buffer's length, the next write throws.

The probe is now installed only for unshared buffers. A shared one leaves it unset, so the fast check fails every time and control falls through to the buffer-identity comparison in `refreshWasmCaches` — which is exactly what the code did before this branch. The unshared path pays nothing: the hot check is unchanged, and the decision is made once per refresh.

The added test reproduces that scenario directly — an allocator whose `realloc` grows a shared memory and returns a pointer past the old length, copying the bytes as a real `realloc` would. With the guard removed it fails with `RangeError: Offset is outside the bounds of the DataView`, the exact failure the review described. Runtime suite is now 161 tests.

### DIFF
--- a/runtime/typescript/src/wire.ts
+++ b/runtime/typescript/src/wire.ts
@@ -422,6 +422,16 @@ export class WireWriter {
   private offset: number;
   private cachedWasmView: DataView | null;
   private cachedWasmBuffer: ArrayBuffer | null;
+  /**
+   * Detachment probe for the two caches above. `allocator.buffer()` reads the
+   * `WebAssembly.Memory.prototype.buffer` *accessor*, and every scalar write
+   * went through it — 21 accessor reads per record on `countActiveUsers`.
+   * Growing wasm memory detaches the old buffer, so a cached `Uint8Array` over
+   * it reports `byteLength === 0`, which is a plain field read. It has to be a
+   * `Uint8Array`: `DataView.prototype.byteLength` throws when detached rather
+   * than reporting 0, so the view cannot probe itself.
+   */
+  private wasmProbe: Uint8Array | null;
 
   constructor(initialSize = 256, allocateLocal = true) {
     const normalizedSize = Math.max(initialSize, 1);
@@ -434,6 +444,7 @@ export class WireWriter {
     this.offset = 0;
     this.cachedWasmView = null;
     this.cachedWasmBuffer = null;
+    this.wasmProbe = null;
   }
 
   static withWasmAllocation(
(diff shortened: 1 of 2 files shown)


engali94 (approved):

engali94:
the ci failure is unrelated and already fixed

jlucaso1:
@engali94 I think we can merge this one

PR #767, opened 2026-07-30
fix(wasm): pass the target's wasm features to wasm-opt

jlucaso1:
`pack wasm` invokes `wasm-opt` with the optimize level and nothing else. Binaryen infers which wasm features a module may use from its `target_features` custom section — and the linker deletes that section under `[profile.release] strip = true`, which is the usual advice for shipping small wasm. `wasm-opt` then rejects the module `pack wasm` just handed it:

```
[wasm-validator error in function 2] unexpected false: memory.copy operations
require bulk memory operations [--enable-bulk-memory-opt], on
(memory.copy ...
```

rustc emits `memory.copy` for `wasm32-unknown-unknown`, and without the section binaryen assumes bulk memory is off.

`optimize.enabled` defaults to true in release, so a project that strips **loses the optimizer entirely**. On the module I measured that is 229 420 bytes shipped where 198 699 was available — 13.4% — and it also means shipping the LEB128 padding `wasm-ld` leaves behind, which is 8 376 bytes of the difference on its own.

**The repo's own example never hit this**: `examples/demo` sets `debug = 1` and does not strip, so its `target_features` section survives and `wasm-opt` is happy.

## The fix

The features are passed explicitly, read from the build rather than hardcoded, so they stay correct if rustc's defaults move. For `wasm32-unknown-unknown` today that is `bulk-memory`, `multivalue`, `mutable-globals`, `nontrapping-fptoint`, `reference-types`, `sign-ext`.

**The query goes through cargo**, not a bare `rustc`. Review caught this and was right: a standalone `rustc --print cfg` reports only the default toolchain's defaults — no `RUSTFLAGS`, no `CARGO_TARGET_*_RUSTFLAGS`, no `.cargo/config.toml`, and not the toolchain a `rust-toolchain.toml` selects. A project enabling `simd128` that way would still have its stripped module rejected, which is the failure this change exists to remove. So it runs `cargo rustc -- --print cfg` with the same target, profile and cargo arguments the build used, and cargo resolves all of it.

Confirmed with `RUSTFLAGS=-C target-feature=+simd128`: the bare rustc query does not report `simd128`, the cargo one does.

Three deliberate choices:

- **Features binaryen does not spell are dropped.** Passing an unknown `--enable-*` makes `wasm-opt` exit, and omitting one is no worse than today's behaviour.
- **When rustc cannot be reached the list is empty**, which leaves the previous behaviour rather than failing a pack over a diagnostic.
- **`bulk-memory` maps to two flags.** Binaryen splits it: `memory.copy` and `memory.fill` sit behind `--enable-bulk-memory-opt`, and rustc emits both, so `--enable-bulk-memory` alone still fails validation.

### Why not `--all-features`

It silences the error and is wrong. It lets `wasm-opt` *emit* SIMD, threads, GC or exception-handling instructions the host engine may not implement, turning a build-time failure into a runtime one. Measured out of curiosity, it also produces a marginally larger module here (198 718 against 198 699).

## Verification

Every link checked rather than argued:

1. `wasm-opt -Os` on the stripped module fails with the error above. Reproduced directly.
2. The same command with the mapped flags succeeds: 229 420 → **198 699 bytes**, and `wasm-tools validate` passes on the result.
3. `pack wasm` **without** this change, on a project with `strip = true`, prints `wasm-validator error … [--enable-bulk-memory-opt]` for functions 2, 4, 5, 6, 7 and fails.
4. `pack wasm` **with** it, same project, prints `✓ Optimizing WASM binary`.

## Testing

Five unit tests: parsing what rustc prints, that `bulk-memory` yields both halves, that an unknown feature is dropped, that the wasm32 triple does enable bulk memory, and that a nonexistent triple yields no flags instead of failing.

`cargo test -p boltffi_cli` 445 pass. `cargo clippy --workspace --all-targets` and `cargo fmt --all` clean.

Note on CI: three `boltffi_macros` `core_only_interned_string_pool` fixture tests fail on `main` as well — run 30531511211 on the `main` branch shows the same three — so they are not from this change.

## Not in this PR

The 8 376 bytes of linker LEB128 padding can also be removed at link time with `-C link-arg=--compress-relocations`, which leaves the instruction stream byte-identical (verified: `wasm-tools print` output matches across 63 817 lines). It needs gating, because `rust-lld` hard-errors with `--compress-relocations is incompatible with output debug information` unless debug info is stripped. Worth its own change, since this one is a correctness fix and that one is a size trade with a condition attached.

### DIFF
--- a/boltffi_cli/src/pack/wasm/mod.rs
+++ b/boltffi_cli/src/pack/wasm/mod.rs
@@ -250,6 +250,70 @@ impl WasmArtifactPath {
     }
 }
 
+/// Binaryen infers which wasm features a module may use from its
+/// `target_features` custom section — and the linker deletes that section under
+/// `[profile.release] strip = true`, which is the usual advice for small wasm.
+/// `wasm-opt` then rejects the module it was just handed: rustc emits
+/// `memory.copy`, and without the section binaryen assumes bulk memory is off.
+///
+/// So the features are passed explicitly, taken from what rustc reports for the
+/// target rather than hardcoded, which keeps them correct if rustc's defaults
+/// move. `--all-features` would also silence the error, but it lets `wasm-opt`
+/// *emit* SIMD, threads or GC instructions the host engine may not implement.
+fn binaryen_feature_flags(triple: &str) -> Vec<&'static str> {
+    rustc_target_features(triple)
+        .iter()
+        .flat_map(|feature| binaryen_flags_for(feature))
+        .copied()
+        .collect()
+}
+
+/// Maps a rustc target feature onto the flags binaryen spells it with. Features
+/// binaryen does not know are dropped: passing an unknown `--enable-*` makes it
+/// exit, and leaving one out is no worse than today.
+fn binaryen_flags_for(feature: &str) -> &'static [&'static str] {
+    match feature {
+        // Binaryen splits bulk memory: `memory.copy` and `memory.fill` sit
+        // behind the `-opt` half, and rustc emits both.
+        "bulk-memory" => &["--enable-bulk-memory", "--enable-bulk-memory-opt"],
+        "multivalue" => &["--enable-multivalue"],
+        "mutable-globals" => &["--enable-mutable-globals"],
+        "nontrapping-fptoint" => &["--enable-nontrapping-float-to-int"],
+        "reference-types" => &["--enable-reference-types"],
+        "sign-ext" => &["--enable-sign-ext"],
+        "simd128" => &["--enable-simd"],
+        "atomics" => &["--enable-threads"],
+        "exception-handling" => &["--enable-exception-handling"],
+        "extended-const" => &["--enable-extended-const"],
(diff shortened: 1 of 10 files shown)


boltffi_cli/src/pack/wasm/mod.rs
```diff
@@ -250,7 +255,99 @@ impl WasmArtifactPath {
...
+        .iter()
+        .flat_map(|feature| binaryen_flags_for(feature))
+        .copied()
+        .collect()
+}
+
+/// Maps a rustc target feature onto the flags binaryen spells it with. Features
+/// binaryen does not know are dropped: passing an unknown `--enable-*` makes it
+/// exit, and leaving one out is no worse than today.
+fn binaryen_flags_for(feature: &str) -> &'static [&'static str] {
```
engali94, line 283: can we update the packaging page in docs/ so it will reflect the setup beign done here

engali94 (approved):

PR #768, opened 2026-07-30
perf(wasm): return the packed empty buffer as a constant

jlucaso1:
`FfiBuf::default().into_packed()` is provably `0`. `Default` forwards to `empty()`, which sets `len: 0`, and `into_packed` opens with `if len == 0 { return 0 }`. But `into_packed` is a real call in the wasm module, so every argument-decode failure site built an `FfiBuf` on the stack and called into it to obtain a constant.

Error paths now return `FfiBuf::EMPTY_PACKED`. The same expression was emitted for both branches of the packed shape, so this also takes the call off the **success** path of every fallible export.

No ABI, API or behaviour change: the value crossing the boundary is identical.

## What it is worth, honestly

**234 bytes on `examples/demo`**, measured by building the wasm target with and without the change (8 925 347 → 8 925 113). The demo has fallible exports across 20 modules. Reproduced twice.

**Nothing measurable on a small surface.** I also measured a workload with a single fallible export: 233 529 bytes either way, delta zero. So the saving scales with the number of fallible exports, and one is not enough to move the module. If you are looking for a headline number this is not it — the argument is that a call which provably returns a constant should be a constant, and 234 bytes on a realistic export surface is the size that follows.

I am flagging this because the effect is easy to fail to reproduce: pick a project with one or two fallible functions and you will see zero, and reasonably conclude the change does nothing.

## Testing

- `cargo test --workspace` — 11 groups, no failures.
- One test added, pinning the premise the constant rests on: an empty buffer has `len == 0`. The other half — `into_packed` returning `0` for a zero length — is `wasm32`-gated and cannot be asserted from a host test, and CI does not run `cargo test --target wasm32`; it is two visible lines in `into_packed`. I would rather say that than add a `#[cfg(target_arch = "wasm32")]` test that never runs and looks like coverage.
- `cargo clippy --workspace --all-targets` and `cargo fmt --all` clean.

Three `boltffi_macros` `core_only_interned_string_pool` fixture tests fail on `main` too (run 30531511211 on `main` shows the same three), so they are not from this change.

### DIFF
--- a/boltffi_core/src/types/buf.rs
+++ b/boltffi_core/src/types/buf.rs
@@ -145,10 +145,19 @@ pub extern "C" fn boltffi_buf_with_len(len: usize) -> FfiBuf {
 
 #[cfg(target_arch = "wasm32")]
 impl FfiBuf {
+    /// Packed form of an empty buffer.
+    ///
+    /// `into_packed` returns `0` whenever `len == 0`, and `FfiBuf::empty()` has
+    /// `len == 0`, so `FfiBuf::default().into_packed()` is always this value.
+    /// Error paths return the constant instead: `into_packed` is a real call in
+    /// the wasm module, so spelling it out made every argument-decode failure
+    /// site build an `FfiBuf` on the stack and call into it.
+    pub const EMPTY_PACKED: u64 = 0;
+
     pub fn into_packed(self) -> u64 {
         let len = self.len;
         if len == 0 {
-            return 0;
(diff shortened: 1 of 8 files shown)


engali94:
this should fix the ci failure: https://github.com/boltffi/boltffi/pull/769

engali94 (approved):

jlucaso1:
> this should fix the ci failure: #769

Thanks

PR #770, opened 2026-07-30
capture: async functions join the per-invocation wrapper ABI

akesson:
Async functions graduate into the per-invocation wrapper ABI: with `BOLTFFI_CAPTURE_WRAPPERS` set, an `#[export] async fn` expands to the same future family the legacy path exports, behind an invocation-minted symbol.

```rust
#[export]
pub async fn shout_later(value: String) -> String { ... }

// expands (gated) to:
pub unsafe extern "C" fn boltffi_inv_mycrate_shout_later_1a2b3c4d(
    value: FfiBuf,                                    // FfiCross-lifted eagerly
) -> RustFutureHandle
// plus boltffi_inv_..._{poll, complete, cancel, free, panic_message}
```

- The record keeps carrying only the start symbol; the poll/complete/cancel/free/panic-message companions are fixed suffixes of it — an ABI convention like the trailing status parameter, not a name derived from user items.
- Arguments lift eagerly at the start call. A malformed buffer sets the last error and returns an already-failed future, so the failure surfaces as `INVALID_ARG` at the completion call — the same channel the legacy async path uses (which records no message at all). Since last-error storage is thread-local, the message is only readable on the start thread; the flip PR's call-sites should read it there.
- `complete(handle, *mut FfiStatus)` lowers the whole value through `FfiCross`, `Result` included: no `PlainComplete`/`FallibleComplete` split, per the #734 contract decision. A non-OK future (invalid arg, cancelled) writes its status and returns the poisoned value.
- Poll/cancel/free/panic-message delegate to the same generic `rustfuture` runtime the legacy exports use; no core or ast changes.

Known parity note: a panic inside the polled future aborts on native today on both the legacy and wrapper paths (only wasm's `poll_sync` catches). Hardening that belongs to the runtime, not this ABI.

### DIFF
--- a/boltffi_macros/src/capture.rs
+++ b/boltffi_macros/src/capture.rs
@@ -264,9 +264,6 @@ fn wrapper_tokens(
     if !item.sig.generics.params.is_empty() {
         return Err("generic functions have no per-invocation wrapper");
     }
-    if item.sig.asyncness.is_some() {
-        return Err("async functions have no per-invocation wrapper yet");
-    }
     if item.sig.variadic.is_some() || item.sig.abi.is_some() {
         return Err("extern or variadic functions have no per-invocation wrapper");
     }
@@ -311,9 +308,20 @@ fn wrapper_tokens(
     let ident = &item.sig.ident;
     let symbol = invocation_symbol(&ident.to_string());
     def.native_symbol = Some(symbol.clone());
-    let symbol_ident = quote::format_ident!("{symbol}");
     let facade = facade();
 
+    if item.sig.asyncness.is_some() {
+        return Ok(async_wrapper_tokens(
+            ident,
+            &symbol,
+            &names,
+            &types,
+            &return_type,
+            &facade,
+        ));
+    }
+
+    let symbol_ident = quote::format_ident!("{symbol}");
     Ok(quote! {
         #[doc(hidden)]
         #[unsafe(no_mangle)]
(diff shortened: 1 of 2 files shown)


engali94:
hey @akesson I was re-studying pr #734 and found something substantial that I missed during the original review.

The new FfiCross path effectively introduces a new ABI classification: primitives cross directly, while almost every other supported value crosses through FfiBuf. That is not how BoltFFI behaves today.

Our classification is the backbone of BoltFFI. We distinguish direct records, encoded records, C-style enums, data enums, direct vectors, scalar options, handles, encoded values, return slots, out-pointers, and surface-specific Native/Wasm shape

The macro and the backend must consume the same classified contract. Neither side should independently decide how a Rust type crosses the boundary.

`FfiCross` introduced a per-invocation wrapper is expanded before the complete source contract has been aggregated and lowered, so the wrapper needs type information locally. But solving that ordering problem by introducing a simpler crossing policy changes BoltFFI’s ABI.
  
For example:

  - Blittable records stop crossing directly and become encoded buffers.
  - C-style enums stop crossing as their integer representation.
  - Direct vectors lose their specialized pointer/length representation.
  - Encoded inputs change from borrowed spans to owned FfiBuf values.
  - Result<T, E> changes from separate success/error channels into one encoded value.
  - Native and Wasm surface-specific decisions are bypassed.

This is precisely where BoltFFI intentionally differs from a uniform-buffer approach such as UniFFI’s. We classify aggressively because we want the crossing to preserve the best representation available for each type and target.

akesson:
@engali94 
You're right and thanks for the feedback!

I propose to grow `FfiCross` into the full classification — blittable records and C-enums are decidable at the `#[data]` site, containers dispatch on their element's mode, borrowed-in/owned-out like `Passable`. To make your principle hold mechanically, records will carry the compiled classification and aggregation will validate it against `ffi_rules`, failing loudly on any mismatch.

One ABI question before code: to keep slot arity fixed at expansion time, encoded named data types would cross as `FfiSpan` by value (the header already reserves it; packed `u64` on wasm). OK with that?

I also noticed that the macro and binding classifiers already disagree today (`ffi_rules` lacks the portable-layout check — `{u32, f64}` is blittable to one side, encoded to the other), so I'm doing a rules-unification PR first.

Holding this PR meanwhile.

akesson:
Rules-unification: #774

PR #774, opened 2026-07-31
fix: classify blittable records through one shared portable-layout rule

akesson:
The workspace has two blittable classifiers that disagree on mixed-alignment records. `boltffi_binding`'s `is_direct` rejects records whose layout differs between native ABI alignment profiles (32-bit x86 aligns `i64`/`u64`/`f64` to four bytes), but `boltffi_ffi_rules::classification::classify_struct` — the classifier behind the macro expansion and the bindgen IR lane — had no such check.

The demo already contains a witness:

```rust
#[data]
pub struct Trade {
    pub id: i64,
    pub symbol_id: i32,
    pub price: f64,
    // ...
}
```

`price` sits at offset 16 under the eight-byte profile and offset 12 under the 32-bit x86 profile, so every shipping backend (which renders from the binding-lowered metadata) declares `Trade` wire-encoded — while the macro expansion classifies the same struct blittable and crosses it as raw struct bytes.

This PR makes the portable-layout check part of the shared rule:

- `classify_struct` now takes `&[Primitive]` (which already knows size and   alignment) and requires layout agreement between the natural and  four-byte-capped alignment profiles; the width-blind `FieldPrimitive`  type is gone. A `classify_struct_fields` helper covers the  not-every-field-is-a-primitive case, so all call sites share the same short-circuit instead of hand-rolling it.
- `boltffi_binding::lower::records::is_direct` delegates to the shared classifier, and `lower/layout.rs::compute` builds its offsets from the same shared walk (`classification::layout_profile`), so the emitted layout cannot drift from what the classifier approved.
- The macro lane and the bindgen IR lane pick up the check through their existing classification calls.

Behavior change: `#[data]` structs whose field offsets or total size differ between alignment profiles (such as `Trade` above) now use wire-encoded transport on the Rust side too, matching what the backends already generate. Bindings and dylib regenerate together, so a rebuild keeps both sides consistent; the flip deserves a release-note line.

A mixed-alignment struct can keep the blittable path by making the padding explicit, so both profiles agree on every offset:

```rust
#[data]
pub struct Trade {
    pub id: i64,
    pub symbol_id: i32,
    _pad: i32,
    pub price: f64,
}
```

## Release-note line

Mixed-alignment `#[data]` structs (field offsets or total size differing between native ABI alignment profiles, e.g. an `i32` followed by an `f64`) now cross wire-encoded on every lane. Previously the Rust wrapper passed them as raw struct bytes while generated backends encoded them. Making the padding explicit with filler fields restores the blittable path.

## Follow-up

The explicit-padding recipe leaks a `_pad` field into every generated foreign API. I'm happy to follow up with a padding-field annotation — skipped in the generated surfaces, zeroed on construction — so mixed-alignment structs stay blittable without the wart. Should I go ahead?

🤖 Generated with [Claude Code](https://claude.com/claude-code) and a human

### DIFF
--- a/boltffi_bindgen/src/ir/definitions.rs
+++ b/boltffi_bindgen/src/ir/definitions.rs
@@ -1,5 +1,5 @@
 use boltffi_ffi_rules::callable::{CallableForm, ExecutionKind};
-use boltffi_ffi_rules::classification::{self, FieldPrimitive, PassableCategory};
+use boltffi_ffi_rules::classification::{self, PassableCategory};
 
 use crate::ir::abi::CallId;
 use crate::ir::ids::{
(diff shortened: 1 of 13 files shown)


engali94:
this is using the to-be deprecated ffi_rules crate, all the languages migrated away except dart from this, the new classification authority is the IR and the AST scanner

akesson:
Missed that the ffi_rules is deprectated. Took the cheeky path instead: #796 with an addition to the ADR instead.

PR #777, opened 2026-08-01
fix(ci): make Windows ARM64 NuGet restore deterministic

### LINKED ISSUE #776: Make the Windows ARM64 NuGet restore check deterministic
The new Windows ARM64 packaging check passed on the PR but failed on `main` because `dotnet publish` only searches the generated local package feed. On a clean runner, `Microsoft.NETCore.App.Host.win-arm64` may not already be installed.

Add NuGet.org as a restore source alongside the local feed so the check consistently verifies that our package selects the `win-arm64` native asset without depending on runner state.

mhedgpeth:
Closes #776.

The Windows ARM64 check only pointed `dotnet restore` at the generated local package feed. That worked when the runner already had the ARM64 host pack and failed on a clean runner.

Add NuGet.org as a second source so the Microsoft host pack can be restored while the demo package still comes from the local feed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -278,14 +278,14 @@ jobs:
       - name: Verify win-arm64 NuGet restore and publish
         shell: pwsh
         run: |
-          $packageDir = Join-Path $PWD 'examples/platforms/csharp/dist/packages'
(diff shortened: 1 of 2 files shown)


mhedgpeth:
@engali94 I'm going to hold off on releasing until `main` is green and this does that.

engali94 (approved):

PR #782, opened 2026-08-03
fix: normalize Windows metadata manifest paths

### LINKED ISSUE #781: fix: normalize Windows paths in metadata builds
I found a Windows compatibility issue in BoltFFI 0.29.1. Metadata builds canonicalize the manifest with std::fs::canonicalize, producing a \\?\\... path that Cargo cannot reliably use for nested workspace members with inherited [workspace.dependencies].

The same manifest works with a normal Windows path. Projects currently need to replace valid workspace = true dependencies with explicit paths as a workaround.

The metadata build should use the same non-verbatim path normalization as the CLI, with a Windows regression test covering a nested workspace and inherited dependency.

mhedgpeth:
## Summary

I found a Windows compatibility issue in metadata builds: `std::fs::canonicalize` produces a `\\?\\` path that Cargo cannot reliably use for nested workspace members with inherited `[workspace.dependencies]`.

This uses `dunce::canonicalize` on Windows, matching the CLI behavior, and adds a regression test covering the affected workspace layout.

I’d like to get this into 0.29.2 once it merges.

Closes #781

### DIFF
--- a/boltffi_bindgen/Cargo.toml
+++ b/boltffi_bindgen/Cargo.toml
@@ -25,6 +25,9 @@ walkdir.workspace = true
 indexmap = "2.7"
 object.workspace = true
 
+[target.'cfg(windows)'.dependencies]
+dunce = "1.0"
+
 [dev-dependencies]
 boltffi_ast.workspace = true
 insta.workspace = true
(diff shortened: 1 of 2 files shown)


mhedgpeth:
@engali94 this came up as a regression when I was working to integrate this into crux, so I'd like to try to release this as `0.29.2`

engali94 (approved):

PR #784, opened 2026-08-03
fix: emit BufFromBytes for the C# wire runtime

### LINKED ISSUE #783: Emit BufFromBytes with the wire runtime
I found a C# generation issue in BoltFFI 0.29.2 while upgrading Crux.

When an exported API uses `Vec<u8>`, the generated `FfiBuf.FromBytes` calls `NativeMethods.BufFromBytes`. The generator currently emits the `boltffi_buf_from_bytes` P/Invoke only when closure parameters require copy-buffer support.

For normal wire-runtime APIs, the generated C# therefore fails to compile:

```text
CS0117: 'NativeMethods' does not contain a definition for 'BufFromBytes'
```

The P/Invoke should be emitted whenever the C# wire runtime is generated, not only when closure parameters require it.

Acceptance criteria:

- a regular exported `Vec<u8>` API emits `NativeMethods.BufFromBytes`
- the generated C# compiles
- regression coverage exercises this case
- the Crux counter packages successfully for Windows x64 and ARM64

mhedgpeth:
## Summary

I found a C# generation bug for regular `Vec<u8>` APIs. The generated wire runtime uses `FfiBuf.FromBytes`, but the `NativeMethods.BufFromBytes` P/Invoke was only emitted when closure parameters required copy-buffer support. That leaves generated C# failing with CS0117.

Emit the P/Invoke whenever the wire runtime is generated, and add renderer coverage plus a .NET compile smoke test for a plain exported `Vec<u8>` API.

The C# smoke helper also isolates `APPDATA` so the compile test does not depend on the host's NuGet configuration.

Closes #783.

### DIFF
--- a/boltffi_backend/src/target/csharp/mod.rs
+++ b/boltffi_backend/src/target/csharp/mod.rs
@@ -231,7 +231,7 @@ impl host::HostBackend for CSharpHost {
     fn assemble<'decl>(
         &self,
         bindings: &Bindings<Self::Surface>,
-        _bridge: &Self::Bridge,
+        bridge: &Self::Bridge,
         _context: &RenderContext<Self::Surface>,
         declarations: Vec<RenderedDeclaration<'decl, Self::Surface>>,
     ) -> Result<GeneratedOutput> {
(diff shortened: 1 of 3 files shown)


mhedgpeth:
@engali94 if you're still ok with this I'm working to get anything I'm finding with crux's test suite worked out, I would love to release this as `0.29.3`

engali94 (approved):

Issue #785, opened 2026-08-04
Release Windows ARM64 CLI binaries


mhedgpeth:
## Summary

BoltFFI can build and package for Windows ARM64, but the release workflow does not publish a Windows ARM64 CLI artifact.

## Current behavior

`.github/workflows/release.yml` includes only:

- `x86_64-pc-windows-msvc` → `boltffi-windows-x86_64`

There is no `aarch64-pc-windows-msvc` release-matrix entry.

The CI workflow already covers Windows ARM64 for the C# dual-architecture NuGet package, so this ticket is specifically about the standalone `boltffi` CLI release asset.

## Requested change

Add a Windows ARM64 build to the release matrix:

- Target: `aarch64-pc-windows-msvc`
- Artifact: `boltffi-windows-arm64.zip`
- Checksum: `boltffi-windows-arm64.zip.sha256`

Verify that the resulting asset is attached to GitHub releases and that the install/download documentation identifies it correctly.

PR #786, opened 2026-08-04
Release Windows ARM64 CLI binary

mhedgpeth:
This PR addresses issue #785 by adding Windows ARM64 binary support for `boltffi_cli`.


mhedgpeth:
@engali94 I'm unblocked on the crux side I believe but realized if we're supporting arm64 as a target we should also support it as a cli target, small issue but for completeness I'm including it. It can go in whatever release we want

engali94 (approved):

PR #795, opened 2026-08-08
fix(kotlin): render asynchronous initializers as suspend factories

benedikt-bartscher:


engali94 (approved):

benedikt-bartscher:
Thanks for the quick review+merge @engali94

PR #806, opened 2026-08-09
perf(wasm): return byte buffers without a length prefix

jlucaso1:
## Problem

A `Vec<u8>` returned from an exported function is framed with a `u32` length prefix. `FfiBuf::wire_encode_owned_bytes` makes room for it by growing the vector and shifting the whole payload forward:

```rust
value.reserve_exact(core::mem::size_of::<u32>());
unsafe {
    value.set_len(byte_count + 4);
    core::ptr::copy(value.as_ptr(), value.as_mut_ptr().add(4), byte_count);
}
```

The caller never needed that prefix. `takePackedWireBytes` reads it and then checks it agrees with the length it already had:

```ts
const payloadLength = this.getView().getUint32(pointer, true);
if (payloadLength !== length - 4 || ...) { throw ... }
```

`length` comes from the packed `u64` the call returned. The prefix could only ever confirm what was already known.

## Precedent

A root return may already differ from the nested wire form. `RootEncoding::Surface` asks the surface for the codec, and `Wasm32::root_string_codec` answers `Utf8String`, which crosses through `FfiBuf::from_vec` with no prefix. Byte buffers had no such variant, so the same generated module does this:

```js
probeReserve → _module.takePackedUtf8String(...)   // String, raw
makeBytes    → _module.takePackedWireBytes(...)    // Vec<u8>, framed
```

This adds `root_bytes_return_codec`, defaulting to the framed `CodecNode::Bytes`, which native keeps. Only wasm asks for the raw shape, so no other target changes. `native_bytes_return_expansion_keeps_wire_framing` pins that.

Owned returns only, matching the existing arm, which takes `TypeExpr::String` and not `TypeExpr::Str`. A borrowed value is written by `borrowed_buffer`, which frames unconditionally, so `&[u8]` and `&'static [u8]` constants keep the framed reader. `only_owned_byte_returns_skip_the_length_prefix` covers all three shapes.

The wasm ABI version moves to 3. The return layout changed, so a host and an artifact built either side of this must not load together: the check at instantiation now reports `ABI version mismatch: expected 3, got 2` instead of handing back a payload with the old prefix on the front.

Parameters are untouched. An argument is written into a fresh allocation, so there is no shift to avoid, and nothing but the buffer carries its length, so the prefix is not redundant there. The argument path keeps `allocWireBytes` and `RootEncoding::Surface`.

## Measurements

Isolating the encoder, native, alternating order, median of 9:

| size | `from_vec` | `wire_encode_owned_bytes` | shift |
|---|---:|---:|---:|
| 4 KB | 28ns | 48ns | 20ns |
| 64 KB | 607ns | 713ns | 106ns |
| 256 KB | 2.4us | 4.7us | 2.3us |
| 1 MB | 11.8us | 23.0us | 11.2us |

Both sides include allocating and filling. At 1 MB the shift costs about as much as building the buffer did.

End to end, both artifacts loaded in one process and measured paired, order flipped each repetition, 21 repetitions, ratio is raw over framed. Both sides are built from this branch, differing only in what `root_bytes_return_codec` answers, so nothing else moves between them:

| size | framed | raw | ratio | p25-p75 | verdict |
|---|---:|---:|---:|---:|---|
| 4 KB | 1260ns | 1133ns | 0.91x | 0.85-1.01 | unresolved |
| 64 KB | 20.2us | 20.1us | 0.99x | 0.90-1.10 | unresolved |
| 256 KB | 28.7us | 27.4us | 0.95x | 0.93-0.99 | faster |
| 1 MB | 119.6us | 107.4us | 0.93x | 0.87-0.98 | faster |

So 5 to 7 percent at 256 KB and above, and nothing this machine can resolve below that. An earlier unpaired run suggested far more at 64 KB; it did not survive pairing, and the honest reading is that the small sizes are noise.

`reserve_exact` does not reallocate here, on the host allocator or under wasm, so this removes a copy rather than an allocation:

```
len=1048576  cap 1048576->1048580  realloc=no
```

## Verification

Correctness over the boundary sizes and the free path:

```
n=0, 1, 3, 4, 4096, 65536, 1048576 all round trip with correct length and contents
5000 repeated 1 KB calls, no corruption
borrowed &[u8] returning b"abcde" arrives as 5 bytes, not 9
```

Mixed versions now stop at instantiation:

```
BoltFFI ABI version mismatch: expected 3, got 2
```

Two tests pinned the old shape. The TypeScript emission assertion was updated, and `wasm_bytes_return_expansion_uses_wire_framing` was renamed to say what it now pins. Three were added: `native_bytes_return_expansion_keeps_wire_framing`, so the framed form cannot be lost from the surfaces that need it; `only_owned_byte_returns_skip_the_length_prefix`, which fails if a borrowed slice or a `&'static [u8]` constant is read unframed; and `wasm_abi_version_is_stable` moved to 3.

`cargo test --workspace` green, `cargo clippy --workspace --all-targets` clean, 166 runtime tests pass.

### DIFF
--- a/boltffi_backend/src/target/typescript/codec/read.rs
+++ b/boltffi_backend/src/target/typescript/codec/read.rs
@@ -19,6 +19,7 @@ pub enum ReadKind {
     String,
     Utf8String,
     Bytes,
+    RawBytes,
     ErrorRecord(RecordId),
     ErrorEnum(EnumId),
 }
@@ -115,6 +116,13 @@ impl ReadExpression {
         }
     }
 
+    fn raw_bytes(expression: Expression) -> Self {
+        Self {
+            expression,
+            kind: Some(ReadKind::RawBytes),
+        }
+    }
+
     fn primitive(primitive: Primitive, expression: Expression) -> Self {
         Self {
             expression,
@@ -194,6 +202,18 @@ impl CodecRead for Reader<'_> {
 
     fn bytes(&mut self) -> Self::Expr {
         Ok(ReadExpression::bytes(Expression::call(
+            Expression::identifier(self.reader.clone()),
+            Identifier::known("readBytes"),
+            ArgumentList::default(),
+        )))
+    }
+
+    /// Only ever the whole of a returned buffer, so there is no reader position
+    /// to advance and nothing to read a length from. The expression here is a
+    /// placeholder: the return path keys off the kind and takes the buffer
+    /// directly.
+    fn raw_bytes(&mut self) -> Self::Expr {
+        Ok(ReadExpression::raw_bytes(Expression::call(
             Expression::identifier(self.reader.clone()),
             Identifier::known("readBytes"),
             ArgumentList::default(),
--- a/boltffi_backend/src/target/typescript/mod.rs
+++ b/boltffi_backend/src/target/typescript/mod.rs
@@ -757,8 +757,10 @@ mod tests {
                 .contents()
                 .contains("const __boltffi_value_allocation = _module.allocWireBytes(value);")
         );
+        // The argument keeps its length prefix; the return does not need one,
+        // since the packed value already carries the length.
         assert!(browser.contents().contains(
-            "return _module.takePackedWireBytes((_exports.boltffi_function_demo_echo_bytes as Function)(__boltffi_value_allocation.ptr, __boltffi_value_allocation.len) as bigint);"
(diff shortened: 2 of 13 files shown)


engali94:
@jlucaso1 thanks for this, how does this affect other languges?

jlucaso1:
> @jlucaso1 thanks for this, how does this affect other languges?

0 effect. Only wasm32 targets (typescript)

engali94 (approved):

PR #817, opened 2026-08-11
fix(scan): resolve re-exported types and restore active-feature detection for the Dart cdylib build

Levyks:
## What broke

`820343e7` ("Promote Binding IR generation", [#808](https://github.com/boltffi/boltffi/pull/808)) broke two things for any crate that splits types across modules and/or gates modules behind a Cargo feature.

### 1. Types referenced through a re-export, not their declaration path, fail to scan

```rust
// src/shapes/mod.rs
mod point;
pub use point::*;   // Point is declared in `point`, re-exported here

#[error]
pub enum ShapeError {
    InvalidOrigin { at: crate::shapes::Point },  // via the re-export, not the declaration path
}
```
Fails with `unsupported source type`, even though this compiles fine as plain Rust. `boltffi_scan`'s qualified-path resolution (`crate::`/absolute paths) only matched a type's exact declaration path — never falling back to chase `pub use` re-export chains the way relative-path resolution already did.

### 2. The Dart cdylib build never learns which Cargo features are active

```rust
#[cfg(feature = "a")]
pub mod a;  // silently dropped from the FFI surface
```
`pack dart` built the actual cdylib with a plain `cargo build`. Cargo only exposes `CARGO_FEATURE_*` to build scripts — never to a normal `rustc` invocation — so the `#[data]`/`#[error]` macros saw no active features. `generate dart` was unaffected (it injects the feature list explicitly), which made this easy to miss in the real packaged build.

Same failure mode already fixed for Python in `8bd6ab4a` — this applies the same fix to Dart.

## `boltffi build` had the same bug

`pack dart` wasn't the only place building this way — the standalone `boltffi build` command (`apple`/`android`/`wasm`/`dart`/`all`) used a separate, still-broken `plain_builder` that constructed the exact same plain-`cargo build` selection. Unified both onto one `expanded_builder` helper so there's a single code path left to get this wrong, and removed `BuildSelection::Package`, which had zero callers left once both were fixed.

## Test plan

- `cargo test -p boltffi_scan --lib` and `cargo test -p boltffi_cli --bin boltffi`
- Three new regression tests, each confirmed to fail against the pre-fix code and pass with the fix: `qualified_path_resolves_type_reexported_by_name`, `dart_cdylib_builds_as_a_binding_expansion`, `build_command_uses_a_binding_expansion`
- Verified against a real multi-module, feature-gated crate: both `pack dart` and `boltffi build dart` now succeed end to end where they previously failed

### DIFF
--- a/boltffi_cli/src/build.rs
+++ b/boltffi_cli/src/build.rs
@@ -136,13 +136,7 @@ pub struct BuildOptions {
 
 #[derive(Clone, Debug, Eq, PartialEq)]
 pub enum BuildSelection {
-    Default {
-        cargo_args: Vec<String>,
-    },
-    Package {
-        package: String,
-        cargo_args: Vec<String>,
-    },
+    Default { cargo_args: Vec<String> },
     Expanded(Box<BindingExpansion>),
 }
 
@@ -241,7 +235,6 @@ impl<'a> Builder<'a> {
     fn package_spec(&self) -> &str {
         match &self.options.selection {
             BuildSelection::Default { .. } => self.config.library_name(),
-            BuildSelection::Package { package, .. } => package,
             BuildSelection::Expanded(expansion) => expansion.package_id(),
         }
     }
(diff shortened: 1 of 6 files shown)


Levyks:
@codex review

engali94:
nice catch @Levyks !

Levyks:
@codex review

engali94 (approved):

PR #819, opened 2026-08-11
fix(wasm): pass --import-undefined so wasm links with streams/async

Levyks:
## Summary

`boltffi_core`'s wasm bridge imports its async/stream wake callbacks (`__boltffi_wake`, `__boltffi_stream_wake`) as plain `extern "C"` functions with no definition anywhere in the crate graph — they're meant to be host-provided wasm imports, resolved at instantiation time, not link time. wasm-ld's default policy treats an unresolved symbol as a hard link error, so `boltffi pack wasm` (and anything that wraps it, e.g. `pack dart-web`) failed to link as soon as a crate actually used streams or async exports:

```
error: rust-lld: error: ...libboltffi_core.rlib(...): undefined symbol: __boltffi_stream_wake
error: rust-lld: error: ...libboltffi_core.rlib(...): undefined symbol: __boltffi_wake
```

The only workaround was manually exporting `CARGO_TARGET_WASM32_UNKNOWN_UNKNOWN_RUSTFLAGS="-C link-args=--import-undefined"` yourself — undocumented, and easy to not know about until you hit the link error.

`Builder::build_wasm_with_triple` now sets that rustflag automatically. It appends to (never overwrites) any existing target-specific or global `RUSTFLAGS` the caller already had set, so this doesn't silently drop flags someone else configured.

## Test plan
- [x] `cargo test -p boltffi_cli` — new unit tests for the flag-composition logic (appends correctly whether nothing, a target-specific value, or a global `RUSTFLAGS` was already present)
- [x] Verified against a real crate (`eletra_protocols`, which uses DLMS/ABNT async streams): `boltffi pack wasm` failed with the undefined-symbol link error before this change and succeeds after it, with no manual env var needed
- [x] `cargo fmt --check`, `cargo clippy --all-targets`

### DIFF
--- a/boltffi_cli/src/build.rs
+++ b/boltffi_cli/src/build.rs
@@ -192,6 +192,7 @@ impl<'a> Builder<'a> {
         command.arg("--target").arg(triple);
 
         self.apply_common_build_args(&mut command);
+        apply_wasm_import_undefined(&mut command, triple);
         command.args(&command_args.command_args);
         self.apply_expansion(&mut command)?;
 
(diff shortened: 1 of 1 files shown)


Levyks:
@codex review

jlucaso1:
Looks good. Thanks for that.

Before merging I would like to propose a narrower fix, because the codebase already contains it. Generated callback imports are declared like this:

```rust
#[link(wasm_import_module = "env")]
unsafe extern "C" { ... }
```

And they link today with no linker flag. boltffi_core is the outlier: `__boltffi_wake` and `__boltffi_stream_wake` are declared in a bare extern "C" block with no module attribute, which is why wasm-ld treats them as undefined symbols rather than imports. Those two are the only bare externs in the crate.

Levyks:
> Looks good. Thanks for that.
> 
> Before merging I would like to propose a narrower fix, because the codebase already contains it. Generated callback imports are declared like this:
> 
> ```rust
> #[link(wasm_import_module = "env")]
> unsafe extern "C" { ... }
> ```
> 
> And they link today with no linker flag. boltffi_core is the outlier: `__boltffi_wake` and `__boltffi_stream_wake` are declared in a bare extern "C" block with no module attribute, which is why wasm-ld treats them as undefined symbols rather than imports. Those two are the only bare externs in the crate.

true, that's a much better fix, and since the approaches are so different, and even the branch name wouldn't make sense anymore with this new approach, i'll close this PR and open a new one linking to issue #820

PR #821, opened 2026-08-11
fix(wasm): declare async/stream wake callbacks as wasm imports

### LINKED ISSUE #820: wasm build fails to link with undefined symbol errors when a crate uses streams or async exports
## Summary

`boltffi pack wasm` fails to link the wasm32 target as soon as the crate being packaged actually exports a stream (`#[ffi_stream]`) or an async function/method. The build fails with:

```
error: rust-lld: error: ...libboltffi_core.rlib(...): undefined symbol: __boltffi_stream_wake
error: rust-lld: error: ...libboltffi_core.rlib(...): undefined symbol: __boltffi_wake
```

## Reproduction

1. Create (or use) a crate with `targets.wasm.enabled = true` that exports at least one `#[export] pub async fn ...` or a stream via `#[ffi_stream]`.

Levyks:
## Summary

Fixes #820.

`__boltffi_wake` and `__boltffi_stream_wake` were the only two bare `unsafe extern "C"` declarations in `boltffi_core` without a `#[link(wasm_import_module = "env")]` attribute — every other generated wasm callback import already carries it (see `boltffi_macros`' callback import codegen). Without it, wasm-ld treats the symbol as a normal undefined reference requiring link-time resolution instead of a host-provided wasm import resolved at instantiation time, so linking failed as soon as a crate exported a stream or an async function.

Adding the attribute these two were missing fixes the link error directly, matching the existing convention already used everywhere else in the crate, instead of working around wasm-ld's default policy from the build tooling side (as #819 did).

This supersedes #819 — same problem, but the fix belongs in `boltffi_core`'s own declarations rather than in `boltffi_cli`'s build invocation, and doesn't need any `RUSTFLAGS`/rustc-argument handling at all. Credit to @jlucaso1 for pointing out `boltffi_core` was the outlier here.

## Test plan
- [x] `cargo test -p boltffi_core --lib`
- [x] `cargo build -p boltffi_core --target wasm32-unknown-unknown`
- [x] `cargo fmt --check`, `cargo clippy --all-targets`
- [x] Verified against a real crate with a fully clean rebuild (`target/` and `dist/wasm/` removed) and no `RUSTFLAGS`/`CARGO_TARGET_*_RUSTFLAGS` set anywhere: `boltffi pack wasm` succeeds

### DIFF
--- a/boltffi_core/src/runtime/future.rs
+++ b/boltffi_core/src/runtime/future.rs
@@ -20,6 +20,7 @@ pub enum WasmPollStatus {
 }
 
 #[cfg(target_arch = "wasm32")]
+#[link(wasm_import_module = "env")]
 unsafe extern "C" {
     fn __boltffi_wake(handle: u32);
 }
--- a/boltffi_core/src/runtime/subscription.rs
+++ b/boltffi_core/src/runtime/subscription.rs
@@ -17,6 +17,7 @@ pub enum StreamPollResult {
 pub type StreamContinuationCallback = extern "C" fn(callback_data: u64, StreamPollResult);
 
 #[cfg(target_arch = "wasm32")]
+#[link(wasm_import_module = "env")]
 unsafe extern "C" {
     fn __boltffi_stream_wake(subscription_handle: u32, result: StreamPollResult);
 }


jlucaso1 (approved):
Nice work <3

jlucaso1:
@engali94 The strange thing is that our CI should have caught this issue. I’ll investigate why that didn't happen and open a follow-up to catch these cases. But the fix is ​​absolutely correct.


edit:

Found why our CI missed this. `rust-toolchain.toml` pins `1.95.0`, and rustc stopped passing `--allow-undefined` on wasm cdylib links somewhere between 1.95 and 1.97. On 1.95 the bare `extern "C"` still resolves as an import, so the bug is invisible here while anyone on current stable hits it.

Same crate, only the toolchain changes:

```rust
// src/lib.rs, crate-type = ["cdylib"]
#[export]
pub async fn answer() -> u32 { 42 }
```

```
RUSTUP_TOOLCHAIN=1.95.0  boltffi build wasm   # Built 1/1 targets successfully
RUSTUP_TOOLCHAIN=1.97.1  boltffi build wasm   # undefined symbol: __boltffi_wake
```

Run it outside the repo, otherwise `rust-toolchain.toml` wins and both pass.

That is also why `examples/demo` never caught it: it exports async and streams and is packed for wasm in CI, but at 1.95.

My follow-up will add the link test plus a CI job on stable, so this class shows up here instead of downstream.

PR #824, opened 2026-08-12
feat(typescript): support cancelling in-flight async calls via AbortSignal

Levyks:
## Summary
- Async calls on the TypeScript/wasm target now accept an optional `{ signal?: AbortSignal; cancelId?: number }` parameter, letting a caller cancel an in-flight call using the standard Web `AbortController`/`AbortSignal` primitives, or a bare numeric id for non-JS frontends (dart-web / KMP) via `__boltffiCancelById`.
- Adds a `throw_when` pre-abort check so an already-aborted signal rejects immediately instead of starting the call.
- Wires the wasm async protocol's existing `cancel` export (previously unused) through `AsyncFutureManager`, which cancels the in-flight future on the Rust side and forces a repoll to deliver the resulting `BoltFFICancelledError`.
- The options bag is read via optional property access rather than a defaulted `{}` object, avoiding a per-call allocation for callers that don't use cancellation. Cancel setup (abort listener + resolve/reject wrappers) only runs when `signal` or `cancelId` is provided.
- Streams are out of scope for this PR; only async functions and methods take `signal` / `cancelId`.

## Test plan
- [x] Built the wasm target and ran the generated bindings against real `AbortController` instances in Node.js, verifying both pre-abort rejection and mid-flight cancellation.
- [x] Verified non-cancelled calls behave unchanged.
- [x] Runtime tests cover `cancelById` and reentrant abort-during-poll.

### DIFF
--- a/boltffi_backend/src/target/typescript/mod.rs
+++ b/boltffi_backend/src/target/typescript/mod.rs
@@ -1276,8 +1276,13 @@ mod tests {
             .expect("browser module");
 
         assert!(browser.contents().contains(
-            "export async function asyncAdd(left: number, right: number): Promise<number>"
+            "export async function asyncAdd(left: number, right: number, options?: { signal?: AbortSignal }, __boltffiCancelId?: number): Promise<number>"
         ));
+        assert!(
+            browser
+                .contents()
+                .contains("const __boltffiSignal = options?.signal;")
+        );
         assert!(
             browser
                 .contents()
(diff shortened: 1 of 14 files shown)


Levyks:
@codex review

Levyks:
@codex review

Levyks:
@jlucaso1 if I understood correctly, you're the maintainer of the WASM bindings here, so you should be the one to look at this?!

I've used this `cancelId` approach instead of using the AbortController more directly to make it easier to extend and reuse this feature in my `dart-web` feature, keeping the number of `wasm (dart) <-> JS`  jumps as small as possible

a future KMP web support would also benefit from this API format when compiling Kotlin to WASM instead of JS directly

jlucaso1:
Design looks right, `AbortSignal` is the correct primitive here.

I checked the part that worried me most, whether a cancel always settles the promise. It does. `poll_sync` short-circuits on `is_cancelled()` before polling, so the deferred repoll can't leave a call hanging. Rechecking `signal?.aborted` after the initial poll is a good catch too, that one is easy to miss.

One thing I'd like to settle before merging: `__boltffiCancelId` ends up in the public signature of every async function and method.

```ts
export async function asyncAdd(left: number, right: number, options?: { signal?: AbortSignal }, __boltffiCancelId?: number): Promise<number>
```

Nothing in tree passes it yet, and once it's in a release it's hard to take back. Can it go inside `options` as `cancelId`? I get the allocation argument, but it only costs an object on calls that actually want cancellation, and it keeps the generated `.d.ts` clean for everyone else. If you'd rather keep it positional, at least drop the `__boltffi` prefix, since it is public API either way.

One perf note, since the suspend path is shared by every async call. Wrapping `resolve`/`reject` and building `detach` unconditionally costs three closures per registered call, so calls that never cancel pay for cancellation. Measured on `pollAsync` alone, paired with alternating order, 31 reps, six runs: the fast path is unchanged, registration roughly doubles, from ~128ns to ~255ns. Slower decisively in every run.

Guarding that block on `if (signal || cancelId !== undefined)` and keeping the raw `resolve`/`reject` otherwise brings it to ~1.13x, which no longer resolves against noise. 41 runtime tests still pass with it.

Two smaller things, not blockers. `cancelIds` is keyed by an id the caller picks: two in-flight calls with the same id and the second `set` silently wins, then whichever settles first deletes the shared key and the other loses cancellation too. Worth documenting the uniqueness contract, since the ids will come from another language's codegen.

And streams don't take a signal. Fine to leave out of this PR, just worth saying so in the description so it doesn't read as covered.

osa8bit (approved):

PR #831, opened 2026-08-12
Add Dart demo pack and tests to CI host-defaults

Levyks:
## Summary

Run the existing Dart demo consumer in the same `demo-platforms` host-defaults job as Java, C#, WASM, and Python.

Dart bindings and `examples/platforms/dart/test/` were already in the tree. Nothing packed them or ran `dart test`, so native Dart codegen never failed CI.

## Changes

- Install Dart 3.10.8 on the host-defaults matrix (Python 3.13).
- Add `dart` to `verify-platform-demos.sh` host defaults.
- Pack Dart for the **current host triple only** (default `pack dart` builds every Dart native target).
- Add `examples/platforms/dart/test-demo.sh` (`dart pub get` + `dart test`).
- Declare `pkgs/demo` as a pub workspace member so generated packages with `resolution: workspace` resolve.

## Expected CI

This should go **red** on current `main` at `dart test` compile time:

1. Async encoded out-pointer returns use `_l$result` before it is declared.
2. `StreamPollResult` is generated as `Int32` but the prelude callback is `Int8`.
3. Stream delivery calls `controller.addAll`, which `StreamController` does not have.

Those are pre-existing generator bugs. This PR only makes CI see them.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -206,6 +206,10 @@ jobs:
         uses: Aandreba/setup-binaryen@v1.0.0
         with:
           token: ${{ github.token }}
+      - if: matrix.python-version == '3.13'
+        uses: dart-lang/setup-dart@v1
+        with:
+          sdk: 3.10.8
       - uses: actions/setup-dotnet@v6
         with:
           dotnet-version: '10.0.x'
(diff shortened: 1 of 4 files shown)


engali94:
hey @Levyks is it possible to resolve conflits?

Levyks:
@engali94 sorry for taking so long, got allocated away from the project that used boltffi into a more urgent one and didn't had the time for it, but now things have calmed down and I plan on fixing it shortly

Levyks:
actually, #847 was already merged and added dart demo to CI, so this PR is no longer necessary

Issue #61, opened 2026-02-19
Add C and C++ support


MuntasirSZN:

Marcuss2:
Sorry for commenting on old issue, but I think this is the best place to place it.

Did a bit of investigation as to what it would entail and there are a lot of unknowns and edge cases.

First off, which version of C++ to support?

There are basically two approaches I see:

- Support C++23 fully and then add toggleable features which allow you to downgrade, e.g. C++23 would use std::exception, but a flag would allow you to downgrade to using exceptions, getting you to C++20 (there std::span is the one thing stoppping us from going down to C++17)
- Implement everything in C++98 and then add toggleable features to upgrade the binding, this would allow adoption basically everywhere at first, at the cost of ergonomics.

There is also the issue of package managers and build systems, since C and C++ don't have a unified system for it.

What do the others think?

MuntasirSZN:
I think the second one is better?

Marcuss2:
I thought about it a bit more. I will likely start with at least somewhat C bindings first.

C++ programs can always use the C bindings, it won't look as nice, but it will be functional.

At least if the current maintainers are looking for help with it.

MuntasirSZN:
C bindings are partial as of now, so maintainers are taking the right path.

Marcuss2:
> C bindings are partial as of now, so maintainers are taking the right path.
> […](#)

The word "partial" is doing a lot of heavy lifting. It is pretty much a raw ABI header (boltffi.h) that serves mainly as the internal foundation other backends are built on. It isn't even registered as a target.

Marcuss2:
Small update: I did a subset of features needed for a project. Would you guys be interested in me upstreaming it as an experimental?

I took the approach of providing a header file, dynamic library and static library and leaving the user with packaging it themself. As there is no standard in C or C++.

Marcuss2:
Opened a first slice for an experimental C target: https://github.com/boltffi/boltffi/pull/837

That PR only registers `Target::C`. The body has the intended surface, deferred work, and a question on whether to land the rest as one PR or as the smaller follow-ups listed there.

engali94:
> Small update: I did a subset of features needed for a project. Would you guys be interested in me upstreaming it as an experimental?
> 
> I took the approach of providing a header file, dynamic library and static library and leaving the user with packaging it themself. As there is no standard in C or C++.

yes I do believe this is the right call for C/C++ 
[...]

PR #837, opened 2026-08-13
feat(bindgen): register the experimental C target

Marcuss2:
This is the first slice of an experimental C target, following [Adding a new target language](https://github.com/boltffi/boltffi/blob/main/docs/contributors/adding-a-target.md). Related: #61.

## This PR

Registers `Target::C` (`name() == "c"`) and fills the exhaustive-match arms so later host, CLI, and pack work can hang off it. Generation still returns `UnsupportedTarget`, and config treats C as disabled. Nothing is user-reachable yet.

## Approach (later PRs)

C consumes the Native surface and calls `CBridge` directly — no extra runtime bridge. The public surface is a package-prefixed, static-inline facade in a single header on top of the existing ABI. Packaging is a header plus host shared/static libraries (`include/` + `lib/`); there is no standard C package format, so we leave the rest of the build system to the consumer.

Intended supported surface for the experimental landing:

- synchronous free functions
- `#[repr(C)]` records
- C-style enums
- constants
- classes as opaque handles
- synchronous callback traits
- typed results

## Proposed follow-up sequence

We already have this work internally and can send it either as one follow-up or as the slices below. Prefer the sliced path unless you would rather review one larger experimental-C PR.

1. **Host scaffold and syntax model** — `boltffi_backend/src/target/c/`, `HostBackend` with unsupported capabilities, C syntax fragments.
2. **Portable C ABI header** — C11 / C++ / MSVC atomics and helpers so the generated header is actually includable.
3. **Ergonomic C facade** — typed wrappers over the ABI (functions, records, enums, classes, callbacks, results). Internally this is one commit because `assemble` / surface / callable / result are coupled; we can try to split it further if you want smaller review units.
4. **CLI and config** — `[targets.c]`, `Experimental::WholeTarget(Target::C)`, `boltffi generate c --experimental`.
5. **Pack** — `boltffi pack c --experimental` stages `{output}/include/<library>.h` and host libraries under `{output}/lib`.
6. **Demo platform and audit** — C11 smoke consumer, `exclude(c, …)` for shapes not supported yet, `just demo-test-audit` green.
7. **Docs** — C page plus configuration / experimental / overview / packaging.

## Deferred

- async, streams, custom types, payload/complex enums, encoded value shapes, inline closures
- a C++-idiomatic target (C++ can consume the C header)
- cross-compilation and a fully validated Windows pack story

## Question for maintainers

Do you want this upstreamed **one PR per step** as above, or as **one larger experimental-C PR** after this registration lands? Happy to file a dedicated tracking issue (rather than continuing on #61) if that is the preferred place for the checklist.

### DIFF
--- a/boltffi_bindgen/src/generate.rs
+++ b/boltffi_bindgen/src/generate.rs
@@ -436,7 +436,7 @@ impl Generation {
             }
             Target::Swift => self.render_swift(),
             Target::TypeScript => self.render_typescript(),
-            Target::Header => Err(GenerationError::UnsupportedTarget { target }),
+            Target::Header | Target::C => Err(GenerationError::UnsupportedTarget { target }),
         }
     }
 
@@ -471,7 +471,7 @@ impl Generation {
             Target::KotlinMultiplatform => self.render_kmp_bindings(bindings),
             Target::CSharp => self.render_csharp_bindings(bindings),
             Target::Dart => self.render_dart_bindings(bindings),
-            Target::Swift | Target::TypeScript | Target::Header => {
+            Target::Swift | Target::TypeScript | Target::Header | Target::C => {
                 Err(GenerationError::UnsupportedTarget { target })
             }
         }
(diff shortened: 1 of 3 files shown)


engali94 (approved):
nice! 🎉

PR #845, opened 2026-08-14
Package C libraries for CMake and pkg-config

Marcuss2:
`pack c` now produces a package that CMake, Make, and Meson projects can consume. It includes relocatable CMake targets for shared and static linking, version metadata, and pkg-config files. Static dependencies come from the Rust build through the native-link parsing already used by JVM packaging. Native filenames are preserved for Windows import libraries, and macOS libraries use an `@rpath` install name.

The documentation covers generation and packaging in Packaging, naming and versions in Configuration, and linking and ownership on the C page. C examples remain alongside the other languages on the API pages. The C demo uses the generated package and exercises both linkage modes after moving it to a directory containing spaces. Linux and macOS CI also exercise pkg-config and Meson consumers.

Validation: 482 CLI tests, strict Clippy, formatting, six CMake/pkg-config consumers and two Meson consumers on macOS, exact-version and pointer-size checks, missing/stale `--no-build` metadata checks, and the documentation build with all 20 Markdown output checks. Static executables were checked for accidental shared Rust-library dependencies. The existing C API examples and resolved review fixes remain in this branch.


engali94 (approved):

engali94:
hey @Marcuss2 thanks for the amazing work
I did some enhancements to the docs, also adding cmake pkg config for `pack c` command

PR #873, opened 2026-08-30
Add targets.android.link.extra_args for custom linker flags

gabbrielisaac:
## Problem

`AndroidPackager::link_shared_library` invokes `clang` directly to link
the final Android `.so`, with linker args fully hardcoded in
`android_shared_link_args`. This means `RUSTFLAGS` and
`.cargo/config.toml` have no effect on this final link step — they only
apply to the earlier `cargo build` that produces the staticlib.

This makes it impossible to add linker flags like
`-Wl,-z,max-page-size=16384` (needed for Android 16k page size support)
without patching the CLI.

## Solution

Adds an optional `[targets.android.link]` config table with an
`extra_args` field, following the same pattern as `AndroidPackConfig`
and `DebugSymbolsConfig`. Extra args are appended after the built-in
flags, so they can override defaults (like page size) when needed.

## Example

\`\`\`toml
[targets.android.link]
extra_args = ["-Wl,-z,max-page-size=16384"]
\`\`\`

## Testing

- `cargo test -p boltffi_cli` (unit tests): 473 passed, 0 failed
- `cargo test -p boltffi_cli --test wasm_wake_link`: 1 pre-existing
  failure unrelated to this change (confirmed via `git stash` — fails
  identically on `main` without this patch)
- Added `android_linker_appends_configured_extra_link_args`
- Added `parses_android_link_extra_args`

## Real-world verification

Verified against a real Android JNI library build using
`extra_args = ["-Wl,-z,max-page-size=16384"]`:

\`\`\`
$ readelf -l lib*.so | grep -A1 LOAD
  LOAD           0x0000000000000000 0x0000000000000000 0x0000000000000000
                 0x00000000000ceda0 0x00000000000ceda0  R      0x4000
  LOAD           0x00000000000ceda0 0x00000000000d2da0 0x00000000000d2da0
                 0x00000000001ae7d0 0x00000000001ae7d0  R E    0x4000
  LOAD           0x000000000027d570 0x0000000000285570 0x0000000000285570
                 0x000000000000ead0 0x000000000000fa90  RW     0x4000
  LOAD           0x000000000028c040 0x0000000000298040 0x0000000000298040
                 0x0000000000000cb0 0x0000000000003890  RW     0x4000
\`\`\`

All LOAD segments are aligned to `0x4000` (16384 bytes), confirming
16KB page size compliance — required for Android 15+ devices.

### DIFF
--- a/boltffi_cli/src/commands/init.rs
+++ b/boltffi_cli/src/commands/init.rs
@@ -2,7 +2,7 @@ use std::path::{Path, PathBuf};
 
 use crate::cli::Result;
 use crate::config::{
-    AndroidConfig, AndroidPackConfig, AppleConfig, CSharpConfig, CargoConfig, Config, DartConfig,
+    AndroidConfig, AndroidLinkConfig, AndroidPackConfig, AppleConfig, CSharpConfig, CargoConfig, Config, DartConfig,
     DebugSymbolsConfig, ErrorStyle, HeaderConfig, JavaConfig, KotlinConfig, KotlinFactoryStyle,
     KotlinMultiplatformConfig, PackageConfig, PythonConfig, SpmConfig, SwiftConfig, TargetsConfig,
     WasmConfig, XcframeworkConfig,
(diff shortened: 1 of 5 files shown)


engali94 (approved):

engali94:
thanks @gabbrielisaac let us make the CI green the rest is all good

gabbrielisaac:
@engali94  Let’s see if we can get the CI green now. Thanks a lot for the quick review and for bearing with me, really appreciate it.

PR #890, opened 2026-09-14
perf(macros): precompute span offsets and declaration ordinals per file

benedikt-bartscher:
Follow-up to #888, which this branch builds on (its two commits are included here).

`from_macro_input` still scans the whole source file twice per `#[data]`: once counting newlines to turn the span's line into a byte offset, and once lexing the file to work out which declaration of that name the span is. Both are per invocation, so a file holding many declarations pays for each one.

Both are now precomputed into the same per-file index: a line-start table, and the byte range of every declaration name in lexical order, keyed by `(name, kind)` — a range's position in that list is the ordinal the scope table is already keyed by. Both are plain offsets and `String`s, so they cache with the scopes and hold nothing the proc-macro bridge owns.

Measured on a crate with ~900 `#[data]` types, release build, three runs:

|        | `macro_expand_crate`     | total build             |
| ------ | ------------------------ | ----------------------- |
| before | 6.6 / 6.6 / 6.8s         | 37.3 / 37.3 / 37.2s     |
| after  | **4.89 / 4.77 / 4.79s**  | **34.1 / 33.4 / 33.7s** |

The generated bindings are byte-identical before and after. `cargo fmt --check`, `cargo clippy -p boltffi_macros --all-targets` and the macro/binding tests are clean.

The test that pinned the old scanning helpers now builds an index and asserts through it, so it covers the path the macro actually takes.


benedikt-bartscher:
lets merge https://github.com/boltffi/boltffi/pull/888 first

engali94:
Nice speedup!

engali94:
hey @benedikt-bartscher 
since https://github.com/boltffi/boltffi/pull/888 is merged should this be ready now?

benedikt-bartscher:
hi @engali94 thanks for the heads-up - merged main and marked pr as ready

engali94 (approved):

PR #891, opened 2026-09-15
feat(pack): select android architectures and skip the desktop natives

benedikt-bartscher:
`pack android` builds every configured architecture and, where the config enables it, the Kotlin desktop natives. Two flags let a caller take a slice of that, so the architectures can be packed as separate parallel CI jobs:

* `--architecture <ARCH>`, repeatable — build only these; an architecture that is not configured is an error rather than an empty package
* `--skip-desktop` — leave the desktop natives alone whatever the config says

Both default to today's behaviour.


boltffi_cli/src/pack/android/mod.rs
```diff
@@ -47,17 +47,21 @@ pub(crate) fn pack_android(
...
+    )?;
 
     let build_cargo_args = resolve_build_cargo_args(config, &options.execution.cargo_args);
     let binding_expansion = (!options.execution.no_build)
         .then(|| BindingExpansion::resolve(config, &build_cargo_args))
         .transpose()?;
     let build_profile =
         crate::build::resolve_build_profile(options.execution.release, &build_cargo_args);
-    let android_targets = config.android_targets();
+    let android_targets = selected_android_targets(config, &options.architectures)?;
```
akesson, line 62: The stale sweep in link.rs still runs against `RustTarget::ALL_ANDROID` and deletes every library it didn't package this run. With arm64 + x86_64 configured, `--architecture arm64` followed by `--architecture x86_64` into the same dist/ leaves only x86_64 in jniLibs.

akesson, line 62: Same for debug symbols: the zip is rebuilt from this run's outputs only, so a per-arch pack overwrites `<crate>.android.symbols.zip` with just that arch.

boltffi_cli/src/pack/android/mod.rs
```diff
@@ -108,6 +112,18 @@ pub(crate) fn pack_android(
         step.finish_success();
     }
 
+    // `--desktop-only` stops here: with no architecture built there is nothing to
+    // discover or package, and the jniLibs already on disk stay as they are.
+    if options.desktop_only {
```
akesson, line 117: When the config doesn't enable bundled desktop natives this exits 0 having built and packaged nothing. Inconsistent with `--architecture`, which errors on an unconfigured value.

boltffi_cli/src/cli.rs
```diff
@@ -273,6 +296,26 @@ pub(crate) enum PackTargetArg {
...
+            conflicts_with = "desktop_only",
+            help = "Skip the Kotlin desktop natives, whatever the configuration says"
+        )]
+        skip_desktop: bool,
+
+        #[arg(
+            long,
+            help = "Build only the Kotlin desktop natives, leaving the Android architectures alone"
+        )]
+        desktop_only: bool,
```
akesson, line 317: `--desktop-only --architecture arm64` is accepted: the selection is validated, then never used. There is no `conflicts_with` between the two.

boltffi_cli/src/pack/android/mod.rs
```diff
@@ -166,8 +186,46 @@ fn ensure_android_kotlin_desktop_no_build_supported(config: &Config, no_build: b
     Ok(())
 }
 
-fn should_package_android_kotlin_desktop_natives(config: &Config) -> bool {
-    config.android_kotlin_desktop_pack_enabled()
+/// The configured Android targets, narrowed to `architectures` when it asks for
+/// some. An architecture the configuration does not carry is an error rather
+/// than an empty build, so a typo cannot quietly produce a partial package.
+fn selected_android_targets(
```
akesson, line 192: No test covers this function or the desktop-only branch.

boltffi_cli/src/pack/android/mod.rs
```diff
@@ -166,8 +186,46 @@ fn ensure_android_kotlin_desktop_no_build_supported(config: &Config, no_build: b
...
+    }
+
+    let unknown: Vec<_> = architectures
+        .iter()
+        .filter(|architecture| {
+            !configured
+                .iter()
+                .any(|target| target.architecture() == **architecture)
+        })
+        .map(|architecture| format!("{architecture:?}").to_lowercase())
```
akesson, line 208: nit: this re-derives what `Architecture::canonical_name()` already returns.

akesson (commented):
packaging.mdx still says pack android builds all ABIs and doesn't mention the new flags. The description says "Two flags" and omits `--desktop-only`.

boltffi_cli/src/cli.rs
```diff
@@ -202,6 +203,28 @@ pub(crate) enum GenerateTargetArg {
     All,
 }
 
+#[derive(Clone, Copy, clap::ValueEnum)]
+pub(crate) enum AndroidArchitectureArg {
```
akesson, line 207: optional: this mirrors `Architecture` one-to-one, plus a `From` impl to keep in sync.

benedikt-bartscher:
@akesson thanks for the review - not sure when i will be able to address it, but i will try my best. if you want to get it in fast feel free to take this PR

Marcuss2:
Small FYI: At Innovatrics, we add this to the internal fork (contains this + some C backend extensions I am trying to upstream) as we need the `--desktop-only` mode. That seems to work well. I agree with the rest of the review comments, but they are not critical for us.

I might pick this up and finish it if I find the time.

